PR #3, opened 2022-10-27
More performant simple lexicase

allan2:
I think this is a bit cleaner and a tad faster.

Between 2.5% and 7.5% in the `simple` benchmark with sample size 100

### DIFF
--- a/src/population.rs
+++ b/src/population.rs
@@ -3,7 +3,7 @@
 #![warn(clippy::unwrap_used)]
 #![warn(clippy::expect_used)]
 
-use std::{borrow::Borrow, mem::swap};
+use std::{borrow::Borrow, mem::swap, cmp::Ordering};
 use itertools::Itertools;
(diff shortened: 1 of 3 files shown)


src/population.rs
```diff
@@ -112,36 +112,48 @@ impl<T> Population<T> {
...
+                let score = ind.scores[idx];
+                match score.cmp(&best_score) {
+                    Ordering::Greater => {
+                        best_score = score;
+                        true
+                    }
+                    Ordering::Equal => true,
+                    Ordering::Less => false,
+                }
+            });
```
JustusFluegel, line 152: Here the following would be more readable if that has no performance impact
```suggestion
            let best_score = *candidates.map(|candidate| candidate.scores[idx]).max().unwrap_or(&0);
            candidates.retain(|candidate| candidate.scores[idx] == &best_score);
```

PR #4, opened 2022-10-27
cargo fmt

allan2:
only change is running `cargo fmt`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "anes"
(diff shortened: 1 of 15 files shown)


NicMcPhee:
@allan2 – So sorry! I totally didn't see this when it was relevant, and now the world has moved on so much that it's no longer gonna work.

I get a ton of noise from GitHub because of how we use it for classes, and I just didn't expect someone to do something so lovely as to create a PR here, so I wasn't paying attention. 😢 Thanks a ton for the contribution, though.

PR #5, opened 2023-09-18
Move error stuff out of the mod.rs

JustusFluegel:
Moved the whole error stuff out of the mod.rs of instruction, instead separated it into multiple files. Also merged RecoverableError and FatalError back together into a single StatefulError to avoid code duplication, but instead added a type parameter Severity and added aliases for FatalError and RecoverableError to StatefulError<_,_,Recoverable/Fatal> respectively.
Also added a IntoState trait for generalizing it a bit more.


NicMcPhee:
Thanks for doing all this!

PR #6, opened 2023-09-23
Tracking issue for Transactional states

JustusFluegel:
Adds the traits for working with different kinds of stacks, and a basic "additional" implementation for a Transaction based on a stack. This implementation, as it is "additional" will store the changes that come in and then apply them to the underlying stack once that is requested. This change also makes some of the method take &mut state instead of taking the state directly where applicable, which opens up the door for some interesting ways for adding instructions that are not based on a huge match statement and for which I may add a pr in the future.

~~I will add a few more comments on the specifics once this is in a more usable state, hopefully until the stream tomorrow.~~ (edit: well this never materialized :joy: )

# Todo-List
## Short-term goals
- [x] Check if using Arrays would be better instead of ~~`MonotonicTuple`~~`ArrayLike`
   - [x] For now maybe just implement ~~`MonotonicTuple`~~`ArrayLike` for Arrays?
   - [x] Choose a better name for ~~`MonotonicTuple`~~`ArrayLike`
- [ ] Transaction implementations
   - [ ] Remove the need for the `Stack` and `StackMut` traits if possible
   - [ ] Work out the traits needed for transactions
       - [ ] "Create Transaction"
       - [ ]  "Transaction"
           - [ ] maybe some kind of "WrapperTransaction" which would include all the ones below which re-implement traits from the stack and are themselves stacks in that way
       - [ ] "Transaction-Try" trait which is implemented for all transactions through a blanket impl and provides a way to work with transactions in closures (`state.transaction_try::<Transaction: WrapperTransaction>(|state| { /* do something */}) -> Result<V, E>`) 
   - [ ] Some kind of general TransactionState wrapper which creates Transactions for stacks used (requested by the `HasStack` and `HasStackMut` trait) 
   - [ ]  Stack Transaction Kinds
       - [ ] "Simple Additive", "Simple Substractive": Some kind of baseline Implementation which just works to compare others to, maybe just copying the stack into a temporary one and applying that.
       - [ ] One or more "intelligent" implementations which may use some kind of Enum, Bitstring or similar to store changes only
       - [ ] "Read-Check-Apply": Provided operations to get stuff are read-only and return references, are actually applied once the transaction returns, with optionally a value to push onto stacks. This will be quite limited and not that useful for more complex operations but should basically create no runtime performance hit, as used stuff is stored via the type system. This may allow for simple instructions like Add, Substract or similar to be faster. Will only implement certain Stack operations
   - [ ] Benchmark the different kinds of transactions
 
## Long-term goals
- [ ] Document the api (mid-term)
- [ ] Create usage examples in a more general documentation
- [ ] Better naming
- [ ] Implement ~~`MonotonicTuple`~~`ArrayLike` for T directly and as such remove the need for seperate .x and .n_x methods (.head and .get_n_head for example would just be combined into .head)
    - Blockers: https://github.com/rust-lang/rust/issues/31844
- [ ] Add a const generic to _n methods for specifying the length of the tuple for consistency with discard
    - Blockers: https://github.com/rust-lang/rust/issues/92827


JustusFluegel:
rebased it on the merged commit from #5

JustusFluegel:
Closed in Favor of #128

PR #10, opened 2023-10-22
Proc macro for push state

### LINKED ISSUE #7: Use the TypeState pattern for the State Builder
Use the typestate pattern for the state builder and move the program into a `.with_programm` to allow setting the max stack size beforehand.

```rs
let stack = State::builder()
    .with_max_stack_size(1000)
    .with_program(p)
    .with_int_input("a",1)
    .build();
```

The typestate pattern should ensure
- Setting a stack limit
- Adding a program
- Adding input values

Maybe also consider adding methods to set the stack size per stack like
```rs
let stack = State::builder()
    .with_max_stack_size_for::<PushInstruction>(1000)
    .with_program(p)
    .with_max_stack_size_for::<PushInteger>(1000)
    .with_int_input("a",1)
    .build();
```
Then ensure setting the stack size for all the stacks used in the builder (only allow calling `.with_program()` or `.with_..._input()` after either setting the global max stack size or the respective max stack size).

JustusFluegel:
Adds a proc macro for creating HasStack & Builder impls for state-like structs.
Closes #7


JustusFluegel:
It may be an idea to add a #[derive(HasStack)] macro as well which will act in the following ways:
- if the derive is before a push_state macro it will derive the HasStack macro whilst ensuring attributes like stack are still remaining, removing from then any options that are for the derive only. It will also add the !has_stack feature to the push_state macro to ensure it isn't generated twice.
- if the derive is after the push_state macro the push_state macro acts as if it wasn't there and doesn't re-emit it as it acts like the has_stack feature was enabled

JustusFluegel:
This should be mostly ready for merge btw

NicMcPhee:
Many @JustusFluegel for all their work on this macro code!

PR #42, opened 2023-12-15
Move lints to a workspace config

### LINKED ISSUE #38: Look for uses of `as` for numeric conversions
We imported the `easy_cast` crate in the `push` package, and that might be useful in other places where I did less than wonderful conversions using `as`.

### LINKED ISSUE #41: Move lints configuration to the workspace
Move lint configuration from the lib.rs of individual packages into the workspace's Cargo.toml. This ensures that we use the same lint configuration everywhere.

JustusFluegel:
Closes #41 

as well as Closes #38 since I added the corresponding lint here and fixed the single instance where it was used.


NicMcPhee:
This is awesome – thanks!

PR #43, opened 2023-12-17
Implement Push programs

NicMcPhee:
This makes several related changes:

- It implements `PushProgram`, which allow for arbitrary nesting of programs
- It implements `PushGene`, which is either `Close` or a `PushInstruction`
   - It also implements generators for Push genes with specified and default density of `Close` genes
- It implements the conversion from a `Plushy` genome (composed of `PushGene`s) to a `PushProgram`
- It updates `PushState` to run `PushProgram`s instead of vectors of `PushInstruction`s.

This is _not_ a tidy refactoring, nor a small easy to read pull request. It ended up touching a _lot_ of files, much of which was necessary. It also encouraged cleaning up a number of organizational issues that aren't strictly part of adding `PushProgram`, but made sense (to me) here.


NicMcPhee:
@JustusFluegel: I don't know if you want to look this over before I merge it in. It's a bit of a mess, TBH, so feel free to focus on more important things (like exams!). 😜

packages/push-macros/src/push_state/printing/generate_builder.rs
```diff
@@ -163,7 +163,7 @@ pub fn generate_builder(
                         ///
                         /// let mut state = PushState::builder()
                         ///     .with_max_stack_size(100)
-                        ///     .with_program([])?
+                        ///     .with_no_program()
```
JustusFluegel, line 166: We could maybe move this into it's own issue

packages/push/src/genome/plushy.rs
```diff
@@ -1,68 +1,134 @@
...
 pub struct Plushy {
-    instructions: Vec<PushInstruction>,
+    genes: Vec<PushGene>,
 }
 
 // TODO: We might want to implement some sort of `Into`
 //   trait instead of just having a getter. Having something
 //   like `to_instructions()` since we're cloning?
 impl Plushy {
+    pub fn new<T: IntoIterator<Item = PushGene>>(iterable: T) -> Self {
```
JustusFluegel, line 77:
```suggestion
    pub fn new(iterable: impl IntoIterator<Item = PushGene>) -> Self {
```
We might as well use  a `impl T` here since Iterator types are hard to specify in general most of the time and should be inferred, as such a generic doesn't really have any advantage

packages/push/src/genome/plushy.rs
```diff
@@ -72,63 +138,66 @@ mod test {
...
-            PushInstruction::IntInstruction(IntInstruction::Multiply),
-            PushInstruction::IntInstruction(IntInstruction::Add),
-            PushInstruction::BoolInstruction(BoolInstruction::And),
-            PushInstruction::BoolInstruction(BoolInstruction::Or),
-            PushInstruction::IntInstruction(IntInstruction::Multiply),
-            PushInstruction::IntInstruction(IntInstruction::Add),
-            PushInstruction::BoolInstruction(BoolInstruction::And),
-            PushInstruction::BoolInstruction(BoolInstruction::Or),
-            PushInstruction::IntInstruction(IntInstruction::Multiply),
+        let parent_genes: Vec<PushGene> = vec![
```
JustusFluegel, line 167:
```suggestion
        let genes: Vec<PushGene> = vec![
```
and then we can just use `genes` and not `genes: parent_genes` below

packages/push/src/instruction/bool.rs
```diff
@@ -22,9 +22,6 @@ pub enum BoolInstruction {
     // BoolFromFloat,
 }
 
-#[derive(thiserror::Error, Debug, Eq, PartialEq)]
-pub enum BoolInstructionError {}
```
JustusFluegel, line 26: Why was this removed?

packages/push/src/instruction/instruction_error.rs
```diff
@@ -0,0 +1,15 @@
...
+#[derive(thiserror::Error, Debug, Eq, PartialEq)]
+pub enum PushInstructionError {
+    // Stack errors can be things like stack over- or underflows.
+    #[error(transparent)]
+    StackError(#[from] StackError),
+    #[error("Exceeded the maximum step limit {step_limit}")]
+    StepLimitExceeded { step_limit: usize },
+    // Int errors can be things like integer overflows.
+    #[error(transparent)]
+    Int(#[from] IntInstructionError),
```
JustusFluegel, line 14:
```suggestion
    /// Stack errors can be things like stack over- or underflows.
    #[error(transparent)]
    StackError(#[from] StackError),
    #[error("Exceeded the maximum step limit {step_limit}")]
    StepLimitExceeded { step_limit: usize },
    /// Int errors can be things like integer overflows.
    #[error(transparent)]
    Int(#[from] IntInstructionError),
```

packages/push/src/instruction/mod.rs
```diff
@@ -1,29 +1,20 @@
 use ordered_float::OrderedFloat;
 
-use crate::{
-    error::{Error, InstructionResult},
-    push_vm::{push_state::PushState, stack::StackError, HasStack},
-};
-use std::{fmt::Debug, fmt::Display, sync::Arc};
+use crate::push::instruction::instruction_error::PushInstructionError;
+use crate::{error::InstructionResult, push_vm::push_state::PushState};
```
JustusFluegel, line 4:
```suggestion
use crate::{push::instruction::instruction_error::PushInstructionError,error::InstructionResult, push_vm::push_state::PushState};
```

JustusFluegel (approved):
LGTM

packages/push/src/instruction/bool.rs
NicMcPhee: Because `BoolInstructionError` would be for specific errors caused by boolean instructions. But all boolean operators (at least that we have at the moment) can never fail.

This is different from `IntInstructionError` since integer instructions can generate errors like overflow.

it's dimly possible that we'll want to bring it back/keep it, but I couldn't think of anything special for booleans that might fail.

packages/push-macros/src/push_state/printing/generate_builder.rs
NicMcPhee: Do you mean the idea of renaming it to `without_program` like suggested in #48? Or something else?

packages/push-macros/src/push_state/printing/generate_builder.rs
```diff
@@ -333,13 +333,14 @@ pub fn generate_builder(
...
             /// - `program` - The program you wish to execute
             #[must_use]
-            pub fn with_program<P>(mut self, program: P)
+            pub fn with_program<P, I>(mut self, program: P)
                 -> ::std::result::Result<#builder_name<#utilities_mod_ident::WithSizeAndData, #(#stack_generics),*>, ::push::push_vm::stack::StackError>
             where
-                P: ::std::iter::IntoIterator<Item = ::push::instruction::PushInstruction>,
+                P: ::std::iter::IntoIterator<Item = I>,
                 <P as ::std::iter::IntoIterator>::IntoIter: ::std::iter::DoubleEndedIterator + ::std::iter::ExactSizeIterator,
+                I: ::std::convert::Into<::push::push_vm::program::PushProgram>
```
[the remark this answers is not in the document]
NicMcPhee: It won't let me apply this suggestion in the web UI because there are deleted lines. I think we'll want a separate commit for this. Should I let this go and we'll fix it in a stream, or should I do the change in VSCode and commit it?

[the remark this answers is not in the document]
JustusFluegel: I don't care tbh, we can do whatever you prefer. All the stuff here was minor anyways.

packages/push/src/genome/plushy.rs
```diff
@@ -72,63 +138,66 @@ mod test {
...
+            BoolInstruction::Or.into(),
+            IntInstruction::Multiply.into(),
+            IntInstruction::Add.into(),
+            BoolInstruction::And.into(),
+            BoolInstruction::Or.into(),
+            IntInstruction::Multiply.into(),
         ];
         let parent = Plushy {
-            instructions: parent_instructions,
+            genes: parent_genes,
```
JustusFluegel, line 190:
```suggestion
            genes,
```

here this was missing from the suggestion above

JustusFluegel (approved):

packages/push-macros/src/push_state/printing/generate_builder.rs
JustusFluegel: nah the changes to the doc comments as they are pretty much unrelated to the changes here

packages/push/src/instruction/bool.rs
JustusFluegel: Ok LGTM

PR #45, opened 2023-12-17
Use valid types from macro input for generated doc examples instead of default int/bool ones

JustusFluegel:
Adds a new sample_values option to stacks to allow providing sample values how they should be used, which will generate doctests for the builder. Also provides a ignore_doctests option to allow ignoring doctests using these values, as those may not actually compile if types are used that are not `pub use`'d for re-import in the doctest.


JustusFluegel:
This is based on #47 so I'd recommend merging that first

NicMcPhee (commented):
This will be a lot easier to "see" when we have #47 merged in, so I may wait until that's happened to make sure I understand what's actually being changed here.

So if I understand this correctly, this allows us to add a `sample_values` entry in doc tests, right? Could you add an example of a doctest that uses that feature? I don't _think_ there is one at the moment, right?

JustusFluegel:
https://github.com/NicMcPhee/rust-ga/pull/45/files#diff-b28ceb3f07375acf1dd6e63ba8956ee7a2f7b4627ee7c76172be460fa65a0e21

JustusFluegel:
`PushState` is using it

JustusFluegel:
> This will be a lot easier to "see" when we have #47 merged in, so I may wait until that's happened to make sure I understand what's actually being changed here.

yep that's what I'd recommend as well

JustusFluegel:
Btw [here](https://github.com/JustusFluegel/rust-ga/compare/fix-format...JustusFluegel:rust-ga:generate-correct-doctests) is a link to the diff between this branch and the formatting one, just to show the actual changes happening

NicMcPhee (approved):
Thanks for making it possible to have examples in doctests that use the macro tools.

PR #47, opened 2024-01-11
fix(fmt): Fix formatting

JustusFluegel:
Error on formatting issues instead of silently ignoring them, also fixes those issues present in the codebase

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,4 +23,9 @@ jobs:
       - run: cargo build --verbose
       - run: cargo test --verbose
       - run: cargo clippy
-      - run: cargo fmt --check -- --config error_on_unformatted=true,error_on_line_overflow=true
+      - run: cargo install rustfmt-unstable
+      # Currently (v1.x) rustfmt accepts unstable features as
+      # arguments only, this wrapper reads the file and passes
+      # them as arguments. Rustfmt v2.x won't have this limitation
+      # anymore and then this wrapper can be removed.
+      - run: rustfmt-unstable --config-file rustfmt.toml -- cargo fmt --check -- --config error_on_line_overflow=true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,4 +23,4 @@ jobs:
       - run: cargo build --verbose
       - run: cargo test --verbose
       - run: cargo clippy
-      - run: cargo fmt --check
+      - run: cargo fmt --check -- --config error_on_unformatted=true,error_on_line_overflow=true
--- a/bacon.toml
+++ b/bacon.toml
@@ -1,0 +1,5 @@
+[jobs.fmt-check]
+command = ["cargo", "fmt", "--check", "--", "--config", "error_on_unformatted=true,error_on_line_overflow=true"]
+
+[jobs.check]
+command = ["cargo", "clippy"]
--- a/bacon.toml
+++ b/bacon.toml
@@ -1,5 +1,0 @@
-[jobs.fmt-check]
-command = ["cargo", "fmt", "--check", "--", "--config", "error_on_unformatted=true,error_on_line_overflow=true"]
-
-[jobs.check]
-command = ["cargo", "clippy"]
--- a/packages/ec-linear/src/mutator/with_rate.rs
+++ b/packages/ec-linear/src/mutator/with_rate.rs
@@ -119,7 +119,6 @@ mod tests {
         let mut rng = rand::thread_rng();
         let num_bits = 100;
         let parent_bits = Bitstring::random(num_bits, &mut rng).unwrap();
-        //  make_random(num_bits, &mut rng);
         let child_bits = mutator.mutate(parent_bits.clone(), &mut rng).unwrap();
 
         let num_differences = zip(parent_bits, child_bits)
--- a/packages/ec-linear/src/mutator/with_rate.rs
+++ b/packages/ec-linear/src/mutator/with_rate.rs
@@ -1,7 +1,7 @@
-use anyhow::Result;
-use ec_core::operator::mutator::Mutator;
 use std::ops::Not;
 
+use anyhow::Result;
+use ec_core::operator::mutator::Mutator;
 use rand::{rngs::ThreadRng, Rng};
 
 use crate::genome::Linear;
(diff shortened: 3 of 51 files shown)


bacon.toml
```diff
@@ -0,0 +1,5 @@
+[jobs.fmt-check]
```
NicMcPhee, line 1: Introducing this file is cool (I like having the formatting check as an option), but it breaks all the other standard `bacon` targets like `test` and `check-all`. Do we need to list them all to have them available?

I definitely wouldn't to merge this PR if it breaks how I typically use `bacon`. Should this be in a separate PR anyway?

.github/workflows/ci.yml
```diff
@@ -23,4 +23,4 @@ jobs:
       - run: cargo build --verbose
       - run: cargo test --verbose
       - run: cargo clippy
-      - run: cargo fmt --check
+      - run: cargo fmt --check -- --config error_on_unformatted=true,error_on_line_overflow=true
```
NicMcPhee, line 26: This is a nice addition – thanks!

Given some of the weird things that happen when it tries to break up long comments (like the big URLs below), I wonder if we want `error_on_unformatted` to be true or not?

I generally agree that long lines are undesirable, but having it reformat comments and strings can be a little odd. Can we just have those be warnings?

packages/ec-linear/src/mutator/with_rate.rs
```diff
@@ -118,7 +118,8 @@ mod tests {
 
         let mut rng = rand::thread_rng();
         let num_bits = 100;
-        let parent_bits = Bitstring::random(num_bits, &mut rng).unwrap(); //  make_random(num_bits, &mut rng);
+        let parent_bits = Bitstring::random(num_bits, &mut rng).unwrap();
+        //  make_random(num_bits, &mut rng);
```
NicMcPhee, line 122:
```suggestion
```

This comment really should been removed – not sure why I left it there.

packages/ec-push-cli/src/main.rs
```diff
@@ -65,12 +65,16 @@ fn main() -> Result<()> {
...
-                    // This will return an error if the program is longer than the allowed max stack size.
+                    // This will return an error if the program is longer than
+                    //  the allowed max stack size.
                     // We arguably should check that and return an error here.
                     .unwrap()
                     .with_int_input("x", input)
                     .build();
-                // This is the degree 3 problem in https://github.com/lspector/Clojush/blob/master/src/clojush/problems/demos/simple_regression.clj
+                // This is the degree 3 problem in
+                // https://github.com/lspector/Clojush/blob/
```
NicMcPhee, line 75: This (and the following reformatting) is problematic because it breaks up these URLs, making it impossible to copy/paste them out of the code to see what's being referred to.

packages/push/examples/complex_regression/main.rs
```diff
@@ -34,7 +34,10 @@ use std::ops::Not;
 
 /*
  * This is an implementation of the "complex regression" problem from the Propeller implementation
- * of PushGP: https://github.com/lspector/propeller/blob/master/src/propeller/problems/complex_regression.cljc
+ * of PushGP:
+ * https://github.com/lspector/propeller/blob/
```
NicMcPhee, line 38: Broke a URL again.

packages/push/src/instruction/bool.rs
```diff
@@ -50,8 +50,10 @@ where
     //   - result goes on the char stack
     let transaction: Transaction<PushState> = state.transaction();
 
-    let string = transaction.try_pop<String>()?; // This version has the transaction be mutable.
-    let (string, transaction) = transaction.try_pop<String>()?; // This version returns a "new" transaction.
+    let string = transaction.try_pop<String>()?;
+    // This version has the transaction be mutable.
+    let (string, transaction) = transaction.try_pop<String>()?;
+    // This version returns a "new" transaction.
```
NicMcPhee, line 56:
```suggestion
    // This version has the transaction be mutable.
    let string = transaction.try_pop<String>()?;
    // This version returns a "new" transaction.
    let (string, transaction) = transaction.try_pop<String>()?;
```

This is all in a big commented out block, so it doesn't _really_ matter, but what `cargo fmt` did was put the comments _after_ the lines they were annotating, which was at best confusing.

NicMcPhee (changes requested):
I really like adding the `cargo fmt --check` as a "real" check to the GitHub actions check – thanks!

The two things I'm concerned about are 

- The changes that get made when it breaks URLs in comments (and, to a lesser degree, other issues with reformatting comments and strings).
- The fact that `bacon.toml` breaks `bacon` for me.

bacon.toml
JustusFluegel: oh yea I could move it into it's own pr, was just there cuz I was using it myself and I forgot to exclude it from the commit. Afaik if there is no bacon config bacon just uses the default one, I don't know if there is some way to fallback to that, so we could either just copy the contents or I'll look into that.

.github/workflows/ci.yml
JustusFluegel: I believe there is a separate limit we can set for comments so we could look into that, but `error_on_unformatted` affects more like spaces left behind on lines that rustfmt for some reason doesn't know how to deal with, so that should probably stay enabled. Cargo fmt doesn't reformat strings and comments by default ( I believe there is some option to enable that tho at least in preview ), all the changes that I applied here were made manually cuz rustfmt can't do them itself, that's why they were never done before as well I believe

packages/push/src/instruction/bool.rs
JustusFluegel: Oh yea this is not cargo fmt's fault but mine xD

.github/workflows/ci.yml
JustusFluegel: Oh error on line overflow also ignores comments & string literals so that should be fine (at least according to the docs )

JustusFluegel: Yea having an option to set it to error on trailing whitespace & warn on everything else would probably be nice

JustusFluegel: we might just want 

```bash
cargo fmt --check -- --config=error_on_line_overflow=true,wrap_comments=true,comment_width=120
```

and then configure the comment width to something higher. This should also ignore any comments containing urls.

NicMcPhee (approved):
Thank you very much for all this work!

PR #50, opened 2024-01-28
Use proper variable name in doctests

JustusFluegel:
All doctests were using `int_stack` for the variable name before, now it uses the correct stack type depending on the stack the tests are generated for.


NicMcPhee (approved):
Thanks for fixing that.

PR #54, opened 2024-01-28
Impl From<VariableName> for PushInstruction and change usages where applicable

JustusFluegel:
Closes #44

PR #56, opened 2024-01-28
Impl `vec_into!`, `arr_into!` macros

### LINKED ISSUE #55: Add vec_into! and arr_into! macros to clean up instruction lists
A vec_into![T; <item list>]; macro (and the corresponding array equivalent) which destructures to T`::from(<item>)` for each item in the list would be nice to simplify especially test cases where we create a instruction list manually

JustusFluegel:
Closes #55

### DIFF
--- a/packages/ec-push-cli/src/main.rs
+++ b/packages/ec-push-cli/src/main.rs
@@ -23,8 +23,9 @@ use ec_core::{
 use ec_linear::mutator::umad::Umad;
 use push::{
     genome::plushy::{GeneGenerator, Plushy},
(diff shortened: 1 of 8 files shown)


JustusFluegel:
based on #54 so merge after that

JustusFluegel:
https://github.com/JustusFluegel/rust-ga/compare/from_variable_name_push_instr...JustusFluegel:rust-ga:vec_into_arr_into diff here

NicMcPhee:
I had had a question about whether these could be functions instead of macros, but that's not feasible. Because we want heterogeneous collections of types that all `.into()` the same type, that can't be done with functions.

PR #63, opened 2024-02-01
Disable umad test for now

### LINKED ISSUE #62: Flaky umad test
Mathematically, we have around a 0.89% chance of the umad test (`genome::plushy:.test::umad`) failing, which results in flaky tests in ci. This results in about a 2.665% percent chance of any of the test runs across the three rust versions failing, which is frankly unacceptable and which I also encountered a few times during creation of pull requests.

We may just want to disable the test for now until we also have a solution for #24

JustusFluegel:
Closes #62


NicMcPhee (approved):
Thanks for doing this. I've been saying we should ignore this test but never getting around to it, and we definitely need to ignore this now that we have CI checks on GitHub.

PR #59, opened 2024-02-01
Allow different exec stack types in push_state macro

JustusFluegel:
Closes #39

PR #66, opened 2024-02-04
Fix wrong merge in #59

JustusFluegel:


NicMcPhee:
Sorry about that. I thought I could quick resolve a merge conflict there and biffed it.

If you're one of my students, _this is why you always consult with other people while handling merge conflicts_!

Thanks for the fix.

PR #67, opened 2024-02-04
Error on warnings in ci

JustusFluegel:
This change errors on any warnings in ci from either clippy, cargo build or cargo test.


JustusFluegel:
Well the ci fails here cuz of #66 but after that it should succeed :LUL:

Issue #52, opened 2024-01-28
Add examples to CI as smoke tests


NicMcPhee:
We don't want to have these run a long time, but we could set them up to do things like 100 individuals for 10 generations just to ensure that everything builds and runs. Maybe this only needs to happen when merging to `main`.

JustusFluegel:
@NicMcPhee let's close this #73 resolved this.

NicMcPhee:
Agreed – thanks for taking care of that and noticing that this can be closed.

PR #73, opened 2024-02-10
Example smoke tests

JustusFluegel:
Runs examples in ci. It may be worth to have  a look at https://crates.io/crates/example-runner-cli to understand the config options.


JustusFluegel:
Oh and for some reason last time I looked crates.io didn't display the readme, so if that is the case just open the repository and there it should be.

JustusFluegel:
Closes #52 btw

NicMcPhee (approved):

PR #76, opened 2024-02-11
Configure vscode to use our rustfmt config

JustusFluegel:
@NicMcPhee make sure your rustfmt-unstable version is recent enough, I only added automatic config detection in the workspace recently, I don't know if it was already present when we tested it. (`cargo install rustfmt-unstable`)


JustusFluegel:
I also felt free to add the same config for helix, the editor I am using personally.

NicMcPhee (approved):

PR #83, opened 2024-02-18
Generalize scorer & training/testing cases

NicMcPhee:
This introduces new `Case` and `Cases` types to represent training and testing cases, with a lot of trait implementations to converting to/from both `Case` and `Cases`.

We also introduced a new `FnScorer` struct that implements `Scorer`, which now has the `Output` type as an associated type instead of a generic.

`GenomeScorer` now takes a `Scorer` instead of a "raw" `Fn` for the scoring function.

These led to a variety in other dependent files and types.

### DIFF
--- a/packages/ec-core/src/individual/ec.rs
+++ b/packages/ec-core/src/individual/ec.rs
@@ -9,7 +9,10 @@ use std::{
 
 use rand::rngs::ThreadRng;
 
-use super::{scorer::Scorer, Individual};
+use super::{
+    scorer::{FnScorer, Scorer},
+    Individual,
+};
 use crate::generator::Generator;
 
 /// `EcIndividual` is a struct that represents an individual in an evolutionary
@@ -93,16 +96,28 @@ impl<GG, S> IndividualGenerator<GG, S> {
 
 /// A trait for adding a scorer to a genome generator, creating
 /// an `IndividualGenerator`.
-pub trait WithScorer<Scorer> {
+pub trait WithScorer {
     /// Add a scorer to the genome generator, creating an `IndividualGenerator`.
-    fn with_scorer(self, scorer: Scorer) -> IndividualGenerator<Self, Scorer>
+    fn with_scorer<S, G, R>(self, scorer: S) -> IndividualGenerator<Self, S>
     where
-        Self: Sized;
+        Self: Sized,
+        S: Scorer<G, R>;
+
+    fn with_scorer_fn<F, G, R>(self, f: F) -> IndividualGenerator<Self, FnScorer<F>>
+    where
+        Self: Sized,
+        F: Fn(&G) -> R,
+    {
+        self.with_scorer(FnScorer(f))
+    }
 }
 
-impl<GG, S> WithScorer<S> for GG {
+impl<GG> WithScorer for GG {
     /// Add a scorer to the genome generator, creating an `IndividualGenerator`.
-    fn with_scorer(self, scorer: S) -> IndividualGenerator<GG, S> {
+    fn with_scorer<S, G, R>(self, scorer: S) -> IndividualGenerator<GG, S>
+    where
+        S: Scorer<G, R>,
+    {
         IndividualGenerator::new(self, scorer)
     }
 }
--- a/packages/ec-core/src/individual/ec.rs
+++ b/packages/ec-core/src/individual/ec.rs
@@ -98,10 +98,10 @@ impl<GG, S> IndividualGenerator<GG, S> {
 /// an `IndividualGenerator`.
 pub trait WithScorer {
     /// Add a scorer to the genome generator, creating an `IndividualGenerator`.
-    fn with_scorer<S, G, R>(self, scorer: S) -> IndividualGenerator<Self, S>
+    fn with_scorer<S, G>(self, scorer: S) -> IndividualGenerator<Self, S>
     where
         Self: Sized,
-        S: Scorer<G, R>;
+        S: Scorer<G>;
 
     fn with_scorer_fn<F, G, R>(self, f: F) -> IndividualGenerator<Self, FnScorer<F>>
     where
@@ -114,9 +114,9 @@ pub trait WithScorer {
 
 impl<GG> WithScorer for GG {
     /// Add a scorer to the genome generator, creating an `IndividualGenerator`.
-    fn with_scorer<S, G, R>(self, scorer: S) -> IndividualGenerator<GG, S>
+    fn with_scorer<S, G>(self, scorer: S) -> IndividualGenerator<GG, S>
     where
-        S: Scorer<G, R>,
+        S: Scorer<G>,
     {
         IndividualGenerator::new(self, scorer)
     }
@@ -125,19 +125,18 @@ impl<GG> WithScorer for GG {
 // G is Genome
 // GG is Genome generator
 // S is Scorer
-// R is the TestResult type
-impl<G, GG, S, R> Generator<EcIndividual<G, R>> for IndividualGenerator<GG, S>
+impl<G, GG, S> Generator<EcIndividual<G, S::Score>> for IndividualGenerator<GG, S>
 where
     GG: Generator<G>,
-    S: Scorer<G, R>,
+    S: Scorer<G>,
 {
     /// Generate a new, random, individual.
     ///
     /// This creates a new genome of type `G` using the genome generator of
     /// type `GG`, and then scores the genome using the scorer of type `S`.
-    /// The genome and the test results (of type `R`) are then
+    /// The genome and the test results (of type `S::Score`) are then
     /// used to create a new `EcIndividual`.
-    fn generate(&self, rng: &mut ThreadRng) -> anyhow::Result<EcIndividual<G, R>> {
+    fn generate(&self, rng: &mut ThreadRng) -> anyhow::Result<EcIndividual<G, S::Score>> {
         let genome = self.genome_generator.generate(rng)?;
         let test_results = self.scorer.score(&genome);
         Ok(EcIndividual::new(genome, test_results))
--- a/packages/ec-ga-cli/src/main.rs
+++ b/packages/ec-ga-cli/src/main.rs
@@ -6,9 +6,9 @@ use anyhow::{ensure, Result};
 use clap::Parser;
 use ec_core::{
     generation::Generation,
-    generator::{collection::CollectionGenerator, Generator},
+    generator::{collection::ConvertToCollectionGenerator, Generator},
     individual::{
-        ec::{self, EcIndividual},
+        ec::{EcIndividual, WithScorer},
         scorer::FnScorer,
     },
     operator::{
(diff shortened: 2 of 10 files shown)


packages/ec-push-cli/src/main.rs
```diff
@@ -121,7 +124,7 @@ fn main() -> Result<()> {
 
     let individual_generator = ec::IndividualGenerator {
         genome_generator: plushy_generator,
-        scorer,
+        scorer: FnScorer(scorer),
```
JustusFluegel, line 127: Same thing here

packages/push/examples/complex_regression/main.rs
```diff
@@ -35,60 +36,91 @@ use crate::args::{Args, RunModel};
...
+}
+
+fn main() -> Result<()> {
     let args = Args::parse();
 
     // Inputs from -4 (inclusive) to 4 (exclusive) in increments of 0.25.
-    let training_cases = (-4 * 4..4 * 4)
-        .map(|n| OrderedFloat(f64::from(n) / 4.0))
-        .collect::<Vec<_>>();
+    let training_inputs = (-4 * 4..4 * 4).map(|n| OrderedFloat(f64::from(n) / 4.0));
```
JustusFluegel, line 104:
```suggestion
    let training_inputs = (-4 * 4..4 * 4).map(|n| Of64::from(n) / 4.0);
```
I am not sure if this works but if it does it should be cleaner

packages/push/examples/complex_regression/main.rs
```diff
@@ -35,60 +36,91 @@ use crate::args::{Args, RunModel};
...
+    let training_cases = Cases::from_inputs(training_inputs, |&i| target_fn(i));
+
+    // The range want is -4 1/8, -3 7/8, -3 5/8, ..., 3 7/8, 4 1/8.
+    // I have to multiply that by 8 to get integer values, so:
+    // -33, -31, -29, ..., 31, 33.
+    let testing_inputs = (-33..=33)
+        .step_by(2)
+        .map(|n| OrderedFloat(f64::from(n) / 8.0));
```
JustusFluegel, line 112: Same thing here

packages/push/examples/complex_regression/main.rs
```diff
@@ -110,7 +142,7 @@ fn main() -> Result<()> {
 
     let population = gene_generator
         .to_collection_generator(args.max_initial_instructions)
-        .with_scorer(&scorer)
+        .with_scorer_fn(&scorer)
```
JustusFluegel, line 145: Same thing as with FnScorer also applies here - I would use `.with_scorer(&scorer)` and wrap `scorer` with an `FnScorer` at initialization

packages/push/src/evaluation/cases.rs
```diff
@@ -0,0 +1,99 @@
...
+    cases: Vec<Case<Input, Output>>,
+}
+
+impl<Input, Output> Cases<Input, Output> {
+    #[must_use]
+    pub const fn new() -> Self {
+        Self { cases: Vec::new() }
+    }
+
+    pub fn from_inputs(
```
JustusFluegel, line 18: Suggestion: Do we also want a `inputs.with_target(|i| -> Output) -> Cases<Input,Output>` method here?

packages/push/src/evaluation/cases.rs
```diff
@@ -0,0 +1,99 @@
...
+                input,
+            })
+            .collect()
+    }
+
+    // TODO: Add `from` that selects randomly from some cases
+    // TODO: Add `from` that gets cases from an external source
+    //    Maybe outside of this type?
+
+    pub fn add_case(&mut self, input: Input, output: Output) {
```
JustusFluegel, line 34: Should this take a `Case` instead of input & output? (or a `impl Into<Case>` then one can pass `(Input, Output)` if needed)

packages/ec-ga-cli/src/main.rs
```diff
@@ -69,7 +72,7 @@ fn main() -> Result<()> {
     };
 
     let individual_generator = ec::IndividualGenerator {
-        scorer,
+        scorer: FnScorer(scorer),
```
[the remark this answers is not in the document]
NicMcPhee: Turns out that this doesn't work (at least not immediately) because there's another use of `scorer` down below (in `make_new_individual` that fails if we turn `scorer` into an `FnScorer`. I think the right answer there is actually to change `.wrap()` so that it takes an `FnScorer`. We should look at that in tomorrow's stream.

packages/ec-push-cli/src/main.rs
NicMcPhee: And the same problem with `wrap()` here as well.

packages/push/examples/complex_regression/main.rs
NicMcPhee: Worked great! Nice catch.

packages/push/examples/complex_regression/main.rs
NicMcPhee: We'll have the same problem with `.wrap()` farther down in the file.

packages/push/src/evaluation/cases.rs
NicMcPhee: Maybe? Would we impl that on `Vec<Input>`?

packages/push/src/evaluation/cases.rs
NicMcPhee: I definitely like that idea. We can do that in tomorrow's stream.

packages/push/src/evaluation/cases.rs
JustusFluegel: on T: IntoIterator<Item = Input> I'd say

packages/ec-core/src/individual/ec.rs
```diff
@@ -93,16 +96,28 @@ impl<GG, S> IndividualGenerator<GG, S> {
...
+    fn with_scorer<S, G, R>(self, scorer: S) -> IndividualGenerator<Self, S>
+    where
+        Self: Sized,
+        S: Scorer<G, R>;
+
+    fn with_scorer_fn<F, G, R>(self, f: F) -> IndividualGenerator<Self, FnScorer<F>>
     where
-        Self: Sized;
+        Self: Sized,
+        F: Fn(&G) -> R,
```
JustusFluegel, line 109:
```rs
fn with_scorer_fn<F, G>(self, f: F) -> IndividualGenerator<Self, FnScorer<F>>
    where
        Self: Sized,
        FnScorer<F>: Scorer<G>,
```

NicMcPhee: We chose to not go this route because (as esitsu@Twitch pointed out) if we replace `F: Fn(&G) -> R` with `FnScorer<F>: Scorer<G>`, someone reading the code doesn't know what they can pass in as `f` without going off and reading `FnScorer` as well. So it would have nicely removed the `R`, but the mental cost of that indirection seemed to outweigh that benefit.

packages/push/src/evaluation/cases.rs
NicMcPhee: Right – my idea was way too specific.

It does sound like a potentially useful thing to add. I'd thought we might get to it in today's stream, but that didn't happen. It's easy enough that I might just do it now, though.

NicMcPhee: Actually, I got stuck here (so much for "It's easy enough" 😜). I tried:

```rust
impl<T, Input> T
where
    T: IntoIterator<Item = Input>,
{
    pub fn with_target<Output>(
        &self,
        target_function: impl Fn(&Input) -> Output,
    ) -> Cases<Input, Output> {
        Cases::from_inputs(self.into_iter(), target_function)
    }
}
```

but it's not happy because "no nominal type found for inherent implementation". Basically, it doesn't like this super general `T` that it doesn't know very much about.

The "suggested" fix is to create a new wrapper type (maybe `Inputs`?) that implements `IntoIterator`, and then add `with_target` to that new type. That seems like a fair bit of work for a function I'm not sure we'll ever use/need.

Thoughts?

JustusFluegel: you just need a trait, you can't impl a generic directly

```rs
trait WithTarget<Input> {
    fn with_target<Output>(self,  target_fn: impl Fn(&Input) -> Output) -> Cases<Input,Output>;
}

impl<T, Input> WithTarget<Input> for T
where
    T: IntoIterator<Item = Input>,
{
    pub fn with_target<Output>(
        self,
        target_function: impl Fn(&Input) -> Output,
    ) -> Cases<Input, Output> {
        Cases::from_inputs(self.into_iter(), target_function)
    }
}
```

packages/ec-core/src/individual/ec.rs
JustusFluegel: :+1:

JustusFluegel:
Actually I left on the last stream while doing some work in the background, and I don't think the move from Scorer<T,R> to Scorer<T, type Score> was a good idea, since it prevents us from implementing different score types for the same input type. Instead, we should have just added a generic to the generator instead. I'll create a pr to merge into this one and you can take a look at that.

JustusFluegel:
https://github.com/NicMcPhee/rust-ga/pull/101 here is the pr

NicMcPhee:
> Actually I left on the last stream while doing some work in the background, and in retrospection I don't think the move from `Scorer<T,R>` to `Scorer<T, type Score>` was a good idea, since it prevents us from implementing different score types for the same input type. Instead, we should have just added a generic to the generator instead. I'll create a pr to merge into this one and you can take a look at that.

I hadn't thought about that limitation, but that definitely could be an issue. Thanks for catching that!

Just to make sure I'm understanding the issue, if we have an associated type, we can only have _one_ associated type for any given "actual" version of the trait. So if our trait is generic in `S`, `T`, and `U`, for a given set of values of `S`, `T`, and `U`, we can only have one associated type. If, on the other hand, our "output" type is also a generic (say `V`), then we can have a variety of output types for the same set of values `S`, `T`, and `U`.

Actually, I don't think that can be true. If it were, then we could only have one kind of `Iterator`, but we can specify `Iterator<Item = …>` and get lots of iterators over different types.

Why can't we have `Scorer<G, Score = …>` to get different kinds of scorers for the same input type `G`?

JustusFluegel:
Actually yes that's true - different Iterators (over different types) are actually implementations on different types, not on the same type

NicMcPhee:
So do we need to move `Score = …` back to being a generic? It does seem like an "output type" (like iterators) so it seems to make sense to me as an associated type.

JustusFluegel:
I don't really think the distinction really should be "output type" vs "input type", it should really more be "do we want to be generic over this type" vs "does this type solely depend on other types" - since we might want to have different kinds of Scores, being generic over the output makes sense to me. See for example also the `Into` trait in the standard library.

NicMcPhee:
That makes sense – thanks.

PR #93, opened 2024-02-25
Cleanup dependencies

NicMcPhee:
This moves `quickcheck` and `proptest` to be `dev-dependencies` (which they should have been all along).

I also modified `.github/dependabot.yml` so that it will (I think) create single larger PRs containing all minor and patch updates instead of filling the PR queue with separate PRs for each update.


.github/dependabot.yml
```diff
@@ -7,6 +7,11 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      dependencies:
+        update-types:
+          - "minor"
+          - "patch"
     ignore:
         # These are peer deps of Cargo and should not be automatically bumped
```
JustusFluegel, line 16:
```suggestion
        # These are peer deps of Cargo and should not be automatically bumped
        - dependency-name: "*"
          update-types: ["version-update:semver-minor", "version-update:semver-patch"]
```

I would go with something like this and outright ignore non-breaking version changes

NicMcPhee: I guess that's not my inclination. I don't want to play whack-a-mole chasing every minor change that comes out, but it's also nice to with some regularity catch up with changes people are making.

My inclination is to wait and see what happens if we check weekly and create single, grouped PRs. If that's too much of a hassle, then we could possible move it to every other week or even monthly.

JustusFluegel: I mean cargo updates to the newest, non-semver breaking version anyways so it's not like we don't get them if we don't update the cargo.toml

JustusFluegel: I'd at least ignore patch updates since those don't include any new features

NicMcPhee: OK, I'm not sure I fully understood what was happening here. I did some reading and I hadn't realized a couple of things:

- Cargo might pull in a higher "than obvious" version (into `Cargo.lock`) if there are multiple versions that are requested and there's a minor version that is high enough to cover all the requests. At the moment, for example, our `Cargo.lock` has `proptest` at 1.4.0 even though what we specify is 1.2.0. I used `cargo tree` to determine that this is because `OrderedFloat` requires 1.4.0, and since 1.2.0 and 1.4.0 share a major version, Cargo just goes with the larger version number.
- `cargo update` will update entries in `Cargo.lock` to the latest minor version (leaving `Cargo.toml` alone). A `--dry-run` on our current version, for example, suggests that it would update `clap` from 4.5.0 to 4.5.1.

So when you say "update the cargo.toml", are you talking about running `cargo update`? Is that something we should do on a semi-regular basis just on principle? If I'm understanding things, doing that would essentially do the same thing as minor and patch updates, without bringing Dependabot into the process. And that (I think) is why you're suggesting ignoring them and only having Dependabot talk to us if there's a major version update (since that might break things).

Am I understanding this correctly? Is there some kind of best practice for running `cargo update` now and then? (A quick search didn't turn up anything obvious.)

JustusFluegel: Basically what I mean. There is no actual need to update the dependencies in Cargo.toml manually for non-semver breaking versions, since cargo will update to those anyways (after updating the lockfile). What might be worth consideration is e.g. adding  a pre-push/ post-pull hook that locally updates the cargo.lock (with cargo update) or something, so that while developing we are testing on the most recent version. The point after all of Cargo.lock is that for example git bisect or ci builds are deterministic and not influenced by external factors, so I think just ensuring the local & remote versions are the same during a push should be enough.

That's why I would ignore the updates from dependabot for patch/minor updates.

NicMcPhee: That all makes sense. I hadn't thought about using a pre-push/post-pull hook to automate `cargo update`, but that would ensure that things are happening consistently. Would that potentially add a lot of time to push/pull actions? Enough that it would be a nuisance for some developers?

JustusFluegel: I don't know, but you would need to enable them manually anways so if it does just don't I guess (?)

JustusFluegel: Maybe I would also add it to pre_commit, since pre_push and post_push are probably ran at a bad time to catch any problems if they happen, since pull probably doesn't get run very often & if push errors that really is an anoyance since you 1) need to amend your previous commit or 2) need to create a new commit - so I'd say make the pre-commit do something like this:

- check if the project builds (cargo test --no-run)
- update if it does (maybe check if the pc has internet connection of some kind beforehand?)
- run the tests afterwards in any case regardless of if the update did happen (cargo test)

tests are currently reasonably fast ( ~1-2 seconds-ish ) so I don't think that would add a lot of friction

NicMcPhee: I like that in principle, but in practice it sounds like a lot to have happen at every commit.

Maybe it's just simpler to have a "business practice" of running `cargo update` every week or two just to keep the little stuff up to date?

JustusFluegel: I don't think it's _that_ heavy tbh but I see your point as well - I think either way would probably work fine

NicMcPhee: I think the fact that we have to have a special case for when you don't have an Internet connection that has to be checked every time we commit does seem a bit much.

NicMcPhee:
I'm going to close this. The world has moved on a lot, and many of the related issues have been dealt with elsewhere.

PR #94, opened 2024-02-26
Switch from Random trait to rand::Distribution, cleanup usages

### LINKED ISSUE #88: Use `Distribution` trait instead of the `generation()` function
The `sample` method in [the `Distribution` trait](https://docs.rs/rand/latest/rand/distributions/trait.Distribution.html) would basically do what we're doing with `generate()` method, but would then hook us up to all the other fun toys in the `Rng` crate.

JustusFluegel:
Closes #88


JustusFluegel:
We might want to use `rng.sample(distr)` instead of `distr.sample(rng)`  just for clarity if it is more readable, in most places it's currently the second option. No actual differences in behavior tho

packages/ec-core/src/generator/mod.rs
```diff
@@ -1,83 +1,2 @@
...
-}
-
-/// Generate a random element from an array of options.
-///
-/// # Errors
-///
-/// This returns an `anyhow::Error` if the array of options
-/// is empty since we can't choose a random element from an
-/// empty array.
-impl<const N: usize, T> Generator<T> for [T; N]
```
NicMcPhee, line 45: I see the value of `SliceCloning` below, but it seems like it still might be useful to be able to create `Distribution`s from arrays or vectors. We can create a `SliceCloning` from a reference to our array or vector (you do that in several places below), but there's definitely some machinery there.

Would it make sense to implement things like `impl<T> From<Vec<T>> for SliceCloning<T>` and similar for arrays? Or even `impl<T> From<IntoIterator<Item = T>> for SliceCloning<T>`?

NicMcPhee (commented):
This is really cool – thanks for all this work! I'm generally very positive about these changes, although I did have a number of comments/questions. 😜

packages/ec-core/src/generator/slice_cloning.rs
```diff
@@ -0,0 +1,40 @@
+use rand::{distributions::Slice, prelude::Distribution};
+
+/// Generate a random element from an array of options, cloning the choosen
+/// element.
+#[derive(Debug, Clone, Copy)]
+pub struct SliceCloning<'a, T>(Slice<'a, T>);
+
+#[derive(Debug, Clone, Copy)]
+pub struct EmptySlice;
```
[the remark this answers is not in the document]
JustusFluegel: Actually that's just more coincidence, I did that to mirror the Error type of the Slice Distribution in the rand crate which is sadly not pub (why?) (maybe I should create a issue on the rand crate) but yea that's probably how to do it. More generally I would also always derive `thiserror::Error` and `miette:Diagnostic`, but this is how `rand` did it so I just copied that for now.

packages/ec-core/src/individual/ec.rs
```diff
@@ -111,20 +110,22 @@ impl<GG, S> WithScorer<S> for GG {
 // GG is Genome generator
 // S is Scorer
 // R is the TestResult type
```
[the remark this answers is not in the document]
JustusFluegel: Yea let's remove the comments. The `...T` was more to avoid name collisions with the corresponding traits. More generally I would say we should just decide on a naming convention for generics in this project, whatever it is in the end (short names, ending them with a `T`, trying to find another descriptive synonym, ...) although I favor something that is more descriptive, so that comments like the ones above are not needed

packages/ec-core/src/individual/ec.rs
```diff
@@ -111,20 +110,22 @@ impl<GG, S> WithScorer<S> for GG {
...
+    GenomeGeneratorT: Distribution<GenomeT>,
+    ScorerT: Scorer<GenomeT, ResultT>,
 {
     /// Generate a new, random, individual.
     ///
     /// This creates a new genome of type `G` using the genome generator of
     /// type `GG`, and then scores the genome using the scorer of type `S`.
     /// The genome and the test results (of type `R`) are then
     /// used to create a new `EcIndividual`.
-    fn generate(&self, rng: &mut ThreadRng) -> anyhow::Result<EcIndividual<G, R>> {
```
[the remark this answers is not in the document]
JustusFluegel: I think in general I agree with the `rand` crate here that the random generation itself should not fail, and the error checking should be moved to the Distribution creation itself. I would think that that covers almost all usecases, apart from something like fetching new values from a api? Although I would say that is not really a Distribution then and should probably be handled in some other way. Can you think of anything where that could be an issue?

packages/ec-core/src/population.rs
```diff
@@ -20,20 +20,19 @@ impl<I> Population for Vec<I> {
 mod generator_trait_tests {
```
JustusFluegel, line 20:
```suggestion
mod tests {
```

packages/push/src/genome/plushy.rs
```diff
@@ -33,35 +36,45 @@ impl std::fmt::Debug for PushGene {
 }
 
 #[derive(Debug, Clone)]
-pub struct GeneGenerator {
+pub struct GeneGenerator<'a> {
     close_probability: f32,
-    instructions: Vec<PushInstruction>,
+    instructions: SliceCloning<'a, PushInstruction>,
```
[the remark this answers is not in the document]
JustusFluegel: Maybe rename the field to `instruction_distribution`? The advantage of having a SliceCloning here is that that moves the checking for empty distributions out of the GeneGenerator. I would maybe be even more general here and do a GeneGenerator<T> where T: Distribution<PushInstruction> and then `instructions_distribution: T`, which would allow any distribution. Any thoughts?

packages/ec-core/src/generator/slice_cloning.rs
```diff
@@ -0,0 +1,40 @@
+use rand::{distributions::Slice, prelude::Distribution};
+
+/// Generate a random element from an array of options, cloning the choosen
+/// element.
+#[derive(Debug, Clone, Copy)]
+pub struct SliceCloning<'a, T>(Slice<'a, T>);
```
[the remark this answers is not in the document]
JustusFluegel: Sure. SliceCloning mirrors `rand::Slice` which is  a Distribution<&'a T>. We can't really do anything else here either since the orphan rules prevent us from just implementing Distribution for slices /vecs/... since the trait is from a foreign crate, so we need some newtype of some kind.

packages/ec-core/src/generator/mod.rs
JustusFluegel: We would need some other kind of newtype since we have dangling references (which the compiler will not allow) if we try to convert a `Vec<T>` -> `SliceCloning<T>` since the SliceCloning doesn't own the actual data, and only references it. But that would probably just be another argument to move the GeneGenerator impl to be more general and allow all kinds of distributions.

JustusFluegel: But that would probably be a reasonable addition, maybe along a `IntoDistribution` and / or `AsDistribution` Trait along the lines of 
```rs
trait IntoDistribution<T> {
    type Distribution: Distribution<T>;
    type Error;

    fn into_distribution(self) -> Result<Self::Distribution,Self::Error>;
}

trait AsDistribution<'a, T> {
    type Distribution: Distribution<T>;
    type Error;

    fn as_distribution(&'a self) -> Result<Self::Distribution, Self::Error>;
}

impl<T> IntoDistribution<T> for Vec<T> {
     type Distribution = VecDistr<T>;
     type Error = EmptySlice;
     
     fn into_distribution(self) -> Result<Self::Distribution,Self::Error> {
         VecDistr::new(self)
     }
}

impl<T> AsDistribution<T> for Vec<T>
where T: Clone {
    type Distribution = SliceCloning<'a, T>;
    type Error = EmptySlice;
     
     fn as_distribution(&'a self) -> Result<Self::Distribution,Self::Error> {
         SliceCloning::new(self)
     }
}

impl<'a, T> AsDistribution<&'a T> for Vec<T> {
    type Distribution = Slice<'a, T>;
    type Error = EmptySlice;
     
     fn as_distribution(&self) -> Result<Self::Distribution,Self::Error> {
         Slice::new(self).map_err(|_| EmptySlice)
     }
}
```

JustusFluegel: We could then even take a `impl IntoDistribution<...>` in for example GeneGenerator, but that would need us handling the error there to, so I would probably just leave it as `impl Distribution<...>` and call that manually where we create it, that also allows us to just `?` it and doesn't require us handling that (unrelated) error in GeneGenerator itsself.

packages/ec-core/src/individual/ec.rs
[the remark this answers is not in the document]
JustusFluegel: See discord for more context

packages/ec-core/src/individual/ec.rs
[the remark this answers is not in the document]
JustusFluegel: See discord for more context

packages/ec-core/src/individual/ec.rs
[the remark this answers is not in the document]
NicMcPhee: Not really, TBH, and I definitely prefer checking once at construction rather than repeatedly at generation. So I'm totally fine with this change.

packages/ec-core/src/generator/mod.rs
JustusFluegel: I added these traits, maybe that also makes the above comments less relevant.

packages/ec-core/src/generator/mod.rs
NicMcPhee, line 1: Is `generator` the right module name anymore? We no longer have a `Generator` trait, so maybe this should be renamed. I'm not sure what the new name would be, though. `distributions` would "make sense", but then it clashes with `rand::distributions`. Maybe that's good (they say similar things), but it also potentially complicates some of the references.

packages/ec-core/src/generator/one_of_macro.rs
NicMcPhee, line 1: This is a nice macro. It would be nice to have some tests on it, which should be do-able since this generates a `Distribution` and as you say there should be an "easy" way to generate an equivalent `Distribution` without using the macro.

packages/ec-core/src/generator/wrappers/owned.rs
```diff
@@ -0,0 +1,57 @@
...
+            0 => Err(EmptySlice),
+            len => Ok(Self {
+                collection: val,
+                range: Uniform::new(0, len),
+                _p: PhantomData,
+            }),
+        }
+    }
+
+    pub fn choices(&self) -> NonZeroUsize {
```
NicMcPhee, line 36: `choices()` seems an odd name to me. I would have thought it would give me some collection of options (choices), but it returns a number. Should it instead be `num_choices()`? Or just `len()`?

packages/ec-core/src/generator/wrappers/owned.rs
```diff
@@ -0,0 +1,57 @@
...
+        // Safety: at construction time, it was ensured that the slice was
+        // non-empty, as such the len is > 0.
+        unsafe { NonZeroUsize::new_unchecked(self.collection.borrow().len()) }
```
NicMcPhee, line 39: I think this is the first and only use of `unsafe` in the project so far. I wasn't sure how I felt about that at first, but it does seem reasonable after doing some homework.

packages/ec-core/src/generator/wrappers/owned.rs
```diff
@@ -0,0 +1,57 @@
...
+#[derive(Debug, PartialEq)]
+pub struct OneOfCloning<T, U> {
+    collection: T,
+    range: Uniform<usize>,
+    _p: PhantomData<U>,
+}
+
+impl<T, U> OneOfCloning<T, U>
+where
+    T: Borrow<[U]>,
```
NicMcPhee, line 16: Just so I'm understanding, this says `T` can be borrowed as a slice of `U`, right? So we can pass in things like `Vec<U>` or array of `U` for `T`.

In a lot of other cases we've done things like say that we have to be able to convert `T` into an iterator over `U`, but here we don't want to do this because we need to store the collection in the instance of `OneOfCloning` and we don't want to have to iterate over `T` and make a new collection to store. Correct?

packages/ec-core/src/generator/wrappers/owned.rs
```diff
@@ -0,0 +1,57 @@
...
+    U: 'a + Clone,
+{
+    fn sample<R: rand::Rng + ?Sized>(&self, rng: &mut R) -> U {
+        let idx = self.range.sample(rng);
+
+        // Safety: at construction time, it was ensured that the slice was
+        // non-empty, and that the `Uniform` range produces values in range
+        // for the slice
+        let slice = self.collection.borrow();
+        unsafe { slice.get_unchecked(idx) }.clone()
```
NicMcPhee, line 55: Now another `unsafe`!

TBH, I'd be inclined to actually do all this with safe code first, and then see if there's actually a performance issue here worthy of the introduction of `unsafe` code. I can actually imagine that this might pay off since we're probably sampling this kind of distribution a fair bit, but I don't _know_ that, and I think I'd like to avoid introducing unsafe code unless there's a demonstrable reason for it.

NicMcPhee (changes requested):
This is really nice and I appreciate all the (probably kind of tedious) work!

Two concerns:

- I feel that `generator` may not be a great name for the module anymore.
- I'm not convinced we want to bring in `unsafe` code without some evidence that the performance improvements it provides are meaningful.

This latter question probably highlights the need for some coherent performance testing on all this. I've been meaning to try to introduce some coherent Criterion benchmarks but it's never come to the top of the pile.

packages/ec-core/src/generator/wrappers/owned.rs
JustusFluegel: yes exactly

packages/ec-core/src/generator/wrappers/owned.rs
JustusFluegel: and I think all the unsafes we use here are really easy to wrap your head around (no raw pointers or something like that) so I think that should be fine. If you wish to keep the ec crate clean of unsafe code tho, we could move the distribution module to a choice_distributions crate if you want to

packages/ec-core/src/generator/wrappers/owned.rs
JustusFluegel: I got the same comment in https://github.com/rust-random/rand/pull/1402 and I changed it to num_choices there so let's do that here as well.

packages/ec-core/src/generator/one_of_macro.rs
JustusFluegel: sure I can do that

packages/ec-core/src/generator/mod.rs
JustusFluegel: I'll rename it to distributions, I think that makes sense

packages/ec-core/src/generator/wrappers/owned.rs
NicMcPhee: I'll need to think about that some. I agree that adding an `.unwrap()` isn't pretty, but there is a part of me that would kinda like having this project free of unsafe code, though.

JustusFluegel: I can add extensive testing to all the unsafe parts and we can run miri in ci if that helps

packages/ec-core/src/distributions/one_of_macro.rs
```diff
@@ -0,0 +1,86 @@
+/// Create a Distribution of a specified type, calling `.into()` on all
+/// elements. Similar to the `[...]` syntax of rust optionally allowing to
+/// specify the type using `uniform_distribution_of![<T> ...]`. To sample from
```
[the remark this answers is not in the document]
NicMcPhee: We talked about this at great length on today's Twitch stream, and I was ultimately convinced that it's worth keeping this. The ability to write things like:

```rust
let instruction_set = uniform_distribution_of![<PushInstruction>
    IntInstruction::Add,
    IntInstruction::Subtract,
    BoolInstruction::And,
    FloatInstruction::Multiply,
];
```

is really nice.

I'm a little concerned about the error messages if you have an empty list. You get compile-time errors (good), but the messages aren't super helpful (bad). Not sure there's much we can do about that here, though, if we want to keep the macros simple.

NicMcPhee (approved):
I know there was a _lot_ of discussion about this in the stream today, but I really appreciate the work you've done here – thanks.

I do think it adds complexity (and it's a bummer that the error messages when the list is empty aren't nicer), but I'm sold (for now) on the idea that it adds enough value to justify the complexity.

Issue #65, opened 2024-02-04
Use the new `with_scorer(..)` and `.into_collection_generator(...)`/`.to_collection_generator(...)` over manual struct construction.


JustusFluegel:
We added the above methods in stream (Sun, 04.02.2024) (in #68) but never gotten around to change all the usages apart from in the complex_regression example, so we should change them just to be a bit more explicit/concise.

JustusFluegel:
I believe this should be done after #94 fyi @NicMcPhee

PR #96, opened 2024-02-26
move -cli crates to examples

### LINKED ISSUE #91: Remove the two `-cli` packages
Both of these should be converted into `examples`. The `push-cli` package may actually already basically be there.

JustusFluegel:
Closes #91


JustusFluegel:
This in conjunction with #83 should also Close #65

NicMcPhee (approved):

NicMcPhee:
My sense is that it will be easier to wait on this until we've merged in #83 (and #101 before that) before merging this in. These are great changes, though, and we definitely want to get them merged in soon!

PR #97, opened 2024-02-26
Allow UMAD to add new elements to empty lists

### LINKED ISSUE #78: Allow UMAD to add to an empty program
Because we iterate over the parent program in UMAD, if that is an empty program, the child program is guaranteed to also be empty, which doesn't seem good. It would be good to allow UMAD to add a few instructions to an empty program.

JustusFluegel:
Closes #78

### DIFF
--- a/packages/ec-linear/src/mutator/umad.rs
+++ b/packages/ec-linear/src/mutator/umad.rs
@@ -7,6 +7,7 @@ use crate::genome::Linear;
 pub struct Umad<GeneGenerator> {
     addition_rate: f64,
     deletion_rate: f64,
+    empty_addition_rate: Option<f64>,
     // Provides the generator needed to generate a new, random gene
     // during the addition phase.
     gene_generator: GeneGenerator,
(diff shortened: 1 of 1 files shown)


packages/ec-linear/src/mutator/umad.rs
```diff
@@ -40,6 +69,15 @@ where
     GeneGenerator: Generator<G::Gene>,
 {
     fn mutate(&self, genome: G, rng: &mut ThreadRng) -> anyhow::Result<G> {
+        if genome.size() == 0 {
+            if let Some(addition_rate) = self.empty_addition_rate {
+                return rng
```
NicMcPhee, line 74: This only ever generates one instruction. I actually think we'd probably like this to potentially generate at least a few instructions. A single instruction often won't do anything, and if there's any pressure to move to empty programs (which is probably how we ended up in this situation), a single instruction will probably just get pushed back down to empty pretty quickly.

That said, I'm not sure what a good alternative would be without adding a ton of new parameters. A simple option would be to just have a value that is how many instructions to add. In practice, though, we'd often want that to be a fraction of the maximum number of instructions in a new program, but I don't want to bring that parameter into this. No sure what the best option is.

NicMcPhee:
I'm still not sure if adding a single instruction will actually help, but we don't have any data either way, so I'm inclined to merge this in and move on.

However, we could convert `EmptyAddition` to a `enum` with a variety of options that could be configured at run time, e.g.,

- `EmptyAddition::Match`
- `EmptyAddtion::Given(rate)`
- `EmptyAddition::MutipleMatch(amount)`
- `EmptyAddition::MultipleGiven(rate, amount)`

etc.

PR #98, opened 2024-02-26
Remove unused quickcheck dependency

### LINKED ISSUE #92: Decide between `quickcheck` and `proptest`
I apparently used `quickcheck` in `ec-linear` and `proptest` in `push`. I really shouldn't be using two separate tools that do similar things, so I should probably decide which of these to use and then just stick with it.

From a practical perspective, I think I have more usage or `proptest` in `push`, so it's probably easier to go with `proptest`, but that's not necessarily a great reason to go that direction.

JustusFluegel:
It turned out we never actually ended up using quickcheck (I think, ci will tell) so I just removed it

Closes #92

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,15 +3,6 @@
 version = 3
 
 [[package]]
-name = "aho-corasick"
-version = "1.1.2"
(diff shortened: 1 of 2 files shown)


NicMcPhee:
Good catch!

PR #99, opened 2024-02-26
Remove Self: Sized bound on collection generator to_* methods

### LINKED ISSUE #75: Remove unnecessary `Sized` in `to_` generator methods
In working out the `ConvertToCollectionGenerator` trait and its implementations, at one point we needed `Self` to implement `Sized`. That requirement was dropped in some later revisions, but we didn't removed the `where C: Sized` when that was no longer needed, so we should do that. I think the strategy is to looking for implementations of `ConvertToCollectionGenerator` and remove any `Sized` bounds that got left behind.

JustusFluegel:
Closes #75


NicMcPhee:
Just to make sure I'm understanding, you're pushing the `Sized` bound in to where it's needed (`into_collection_generator`), so the bits that don't actually need it (`to_collection_generator`) aren't limited by it?

JustusFluegel:
yep that's it

PR #100, opened 2024-02-26
Remove old generate trait

JustusFluegel:
The old generate trait was deprecated but never actually removed, so this does that


NicMcPhee:
Thanks for catching this!

PR #101, opened 2024-03-02
Revert move from Scorer<G,R> to Scorer<G, Score = ...>

JustusFluegel:
See #83


NicMcPhee:
I asked a question about this over in #83.

NicMcPhee (approved):
This makes a lot of sense – thanks!

We'll probably start today's stream by merging this in. The one addition I might make is adding some documentation about the use of `PhantomData` in `genome_scorer.rs`. That's likely to be a surprise/confusing for some folks who are coming to this more from the EC side than from the Rust side, so having a comment about it would make sense.

Issue #48, opened 2024-01-11
Naming changes


JustusFluegel:
We've come across a few places where renames might be good during the stream. This issue is for accumulating those and maybe fixing them in the future.

- `max_stack_size` -> `stack_capacity`
- `with_no_program` -> `without_program`

PR #102, opened 2024-03-03
 Gene generator creation

JustusFluegel:
This is meant to implement the `.into_gene_generator(CloseProbability::Uniform)` and `.to_gene_generator(CloseProbability::Given(0.1))` methods, but is currently waiting on https://github.com/rust-random/rand/pull/1402

( oh and as this is based on #94 the changelog is quite messy, see [here](https://github.com/JustusFluegel/rust-ga/compare/rand_distrb_trait...JustusFluegel:rust-ga:gene_generator_creation) for the direct diff )

### DIFF
--- a/packages/ec-core/src/distributions/wrappers/owned.rs
+++ b/packages/ec-core/src/distributions/wrappers/owned.rs
@@ -6,6 +6,14 @@ use crate::distributions::{choices::ChoicesDistribution, wrappers::slice_cloning
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct OneOfCloning<T, U> {
+    // It is really important here that the fields `collection`, `range` and `num_choices` are
+    // never modified, since they all contain information about the length of the collection
+    // which need to be in sync for no panics to occur.
+    //
+    // Therefore, these fields may *never* be pub and no methods may be introduced which can modify
+    // fields without keeping this contract
+    //
+    // Currently these fields are *never* modified at all.
     collection: T,
     range: Uniform<usize>,
     num_choices: NonZeroUsize,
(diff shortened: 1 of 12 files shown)


JustusFluegel:
Ok this is mostly done, now only waiting on rand to release a new version with the `num_choices()` method included

@NicMcPhee you can review this if you feel inclined to, even if it is still a draft (the only changes would be removing the [patch] from Cargo.toml)

JustusFluegel:
Probably also a good idea to change the `.into../.to...` names at some point cuz they are quite long (maybe ask on stream for any better ideas?) but we can always do that in #48 if we want to

JustusFluegel:
The pr is now merged into main in rand, so it's a question of time till it's in the release

JustusFluegel:
Ok rand released `0.9.0-alpha.1` which contains my pr, so we should be good to go ahead and merge this. I think since this is not released anywhere it being an alpha version shouldn't really matter, and I don't think any major api changes will be introduced till the 0.9 release.

packages/ec-core/src/distributions/wrappers/owned.rs
```diff
@@ -2,12 +2,13 @@ use std::{borrow::Borrow, marker::PhantomData, num::NonZeroUsize};
...
 
-use crate::distributions::wrappers::slice_cloning::EmptySlice;
+use crate::distributions::{choices::ChoicesDistribution, wrappers::slice_cloning::EmptySlice};
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Eq)]
 pub struct OneOfCloning<T, U> {
     collection: T,
     range: Uniform<usize>,
+    num_choices: NonZeroUsize,
```
NicMcPhee, line 21: The relationship between these three fields needs to be clearly documented, and we should make sure we have tests (possibly doctests) that confirm that after calling `::new()` the relationships between these fields are preserved.

NicMcPhee (changes requested):
This looks good but we should add tests for the new choices distribution constructor.

NicMcPhee (approved):

PR #109, opened 2024-03-07
Add some new integer predicates

NicMcPhee:
This adds `IsZero`, `IsPositive`, and `IsNegative` to the set of `IntInstruction`s.


JustusFluegel (approved):
LGTM - we might want to move the instructions into their own files in the future tho, but I think we probably should do that in a future pr with all the other instructions moved as well.

PR #110, opened 2024-03-07
Implement smallest

NicMcPhee:
This is a start on the problem Smallest from the PSB1 benchmark suite. It's not really done yet because we don't have the training cases set up correctly and I don't have all the relevant instructions implemented (at a minimum we need `IfElse`).

I'm also not entirely sure where we want things like the PSB1 and PSB2 benchmark implementations to go. I've just stuck this in `examples` with the two regression problems, but I don't think that's where we want these benchmark problems. I'm guessing they should be in a new crate/package, with a separate binary for each benchmark problem. The problem with that is that if someone wants to try a benchmark with a different mutation operator or selection operator (for example), then they'd need to hack that package, which is exactly what we _don't_ want people to be doing. It would be good to set these up in some way where we can inject all the necessary configuration from "the outside", so the implementation of the benchmark problems doesn't need to be tweaked to run them with different (perhaps experimental) tooling.

This also needs #109 to be merged in before it works as is.


packages/push/examples/smallest/args.rs
```diff
@@ -0,0 +1,41 @@
...
+#[derive(clap::ValueEnum, Copy, Clone, Debug)]
+pub enum RunModel {
+    Serial,
+    Parallel,
+}
+
+#[derive(clap::ValueEnum, Copy, Clone, Debug)]
+pub enum TargetProblem {
+    DegreeThree,
+}
```
JustusFluegel, line 12: We should probably remove this like we did in all the other examples

packages/push/examples/smallest/args.rs
```diff
@@ -0,0 +1,41 @@
...
+#[clap(author, version, about, long_about = None)]
+pub struct Args {
+    /// Should we use parallelism when doing the run?
+    #[clap(short, long, value_enum, default_value_t = RunModel::Parallel)]
+    pub run_model: RunModel,
+
+    /// The target problem to run
+    #[clap(short, long, value_enum, default_value_t = TargetProblem::DegreeThree)]
+    pub target_problem: TargetProblem,
+
```
JustusFluegel, line 25: and here then as well

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    push_vm::{program::PushProgram, push_state::PushState, stack::StackError, HasStack, State},
+};
+use rand::{distributions::Distribution, rngs::ThreadRng, thread_rng, RngCore};
+use strum::IntoEnumIterator;
+
+use crate::args::{Args, RunModel};
+
+// An input for this problem is a tuple of four `i64`s.
+#[derive(Copy, Clone)]
+struct Input(i64, i64, i64, i64);
```
JustusFluegel, line 37: do we want to use 
```suggestion
struct Input([i64;4]);
```

here instead?

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+// evolutionary computation. July 2015. ACM.
+//
+// Problem Source: C. Le Goues et al., "The ManyBugs and IntroClass
+// Benchmarks for Automated Repair of C Programs," in IEEE Transactions on
+// Software Engineering, vol. 41, no. 12, pp. 1236-1256, Dec. 1 2015.
+// doi: 10.1109/TSE.2015.2454513
+//
+// This problem is quite easy if you have a `Min` instruction, but can be
+// more a bit more difficult without that instruction.
+fn main() -> Result<()> {
```
JustusFluegel, line 55: I think in the other examples the main is at the bottom - do we want to keep the same ordering?

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    let mut rng = thread_rng();
+
+    let args = Args::parse();
```
JustusFluegel, line 58: In the other examples I de-structure the args - might be preferable here as well

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+        .map(|_| {
+            Input(
+                (rng.next_u32() % 100).into(),
+                (rng.next_u32() % 100).into(),
+                (rng.next_u32() % 100).into(),
+                (rng.next_u32() % 100).into(),
+            )
+        })
+        .collect::<Vec<_>>()
+}
```
JustusFluegel, line 183: I would probably mostly inline this, with something like 

```suggestion
fn random_inputs_iter(rng: &mut ThreadRng) -> impl Iterator<Item = Input> {
        (0..).into_iter().map(|_| {
            Input(
                (rng.next_u32() % 100).into(),
                (rng.next_u32() % 100).into(),
                (rng.next_u32() % 100).into(),
                (rng.next_u32() % 100).into(),
            )
        })
}
```

and then use `random_inputs_iter().take(NUM_TRAINING_CASES)` in the main

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+
+fn smallest(&Input(a, b, c, d): &Input) -> Output {
+    #[allow(clippy::unwrap_used)]
+    Output([a, b, c, d].into_iter().min().unwrap())
+}
+
+fn training_cases(rng: &mut ThreadRng) -> Cases<Input, Output> {
+    training_inputs(rng).with_target_fn(smallest)
+}
```
JustusFluegel, line 192: and this I would probably inline as well, since this is so simple - no real readability advantage here, just needing to jump through a lot of functions

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+
+fn training_inputs(rng: &mut ThreadRng) -> Vec<Input> {
+    const NUM_TRAINING_CASES: usize = 100;
+    (0..NUM_TRAINING_CASES)
+        .map(|_| {
+            Input(
+                (rng.next_u32() % 100).into(),
+                (rng.next_u32() % 100).into(),
+                (rng.next_u32() % 100).into(),
+                (rng.next_u32() % 100).into(),
```
JustusFluegel, line 179: Don't use this here - create a uniform distribution at the top (or inside the .map) - this has 2 advantages:
1. Input generation is faster
2. It doesn't have a bias like the `%` way does

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+// more a bit more difficult without that instruction.
+fn main() -> Result<()> {
+    let mut rng = thread_rng();
+
+    let args = Args::parse();
+
+    let training_cases = training_cases(&mut rng);
+
+    let scorer = FnScorer(
+        |genome: &Plushy| -> TestResults<test_results::Error<i128>> {
```
JustusFluegel, line 63: No need for the return type annotations here ( rustfmt should then reformat this to the following )

```rs
let scorer = FnScorer(|genome: &Plushy| score_genome(genome, &training_cases));
```

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+
+fn score_genome(
+    genome: &Plushy,
+    training_cases: &Cases<Input, Output>,
+) -> TestResults<test_results::Error<i128>> {
+    // The penalty value to use when an evolved program doesn't have an expected
+    // "return" value on the appropriate stack at the end of its execution.
+    const PENALTY_VALUE: i128 = 1_000;
+    let program = Vec::<PushProgram>::from(genome.clone());
+    let errors: TestResults<test_results::Error<i128>> = training_cases
```
JustusFluegel, line 121: just remove the let errors = here and directly return it ( i think this might be due to the switch to i128 i did - if it is sorry I missed this )

packages/push/examples/smallest/main.rs
JustusFluegel: And if we do that, it might just make sense to do `type Input = [i64;4];` instead

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+use rand::{distributions::Distribution, rngs::ThreadRng, thread_rng, RngCore};
+use strum::IntoEnumIterator;
+
+use crate::args::{Args, RunModel};
+
+// An input for this problem is a tuple of four `i64`s.
+#[derive(Copy, Clone)]
+struct Input(i64, i64, i64, i64);
+// An output for this problem is an `i64`.
+struct Output(i64);
```
JustusFluegel, line 39: And then here 
```suggestion
type Output =i64;
```

JustusFluegel: This avoids wrapping in places (like in the score fn below) and should make the code more concise

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    Output([a, b, c, d].into_iter().min().unwrap())
+}
+
+fn training_cases(rng: &mut ThreadRng) -> Cases<Input, Output> {
+    training_inputs(rng).with_target_fn(smallest)
+}
+
+fn instructions() -> Vec<PushInstruction> {
+    let int_instructions = IntInstruction::iter()
+        // Restore this line to remove `Min` from the instruction set.
```
JustusFluegel, line 196: This comment seems quite confusing, since the line below is not commented out

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+        .map(Into::into);
+    let variables = ["a", "b", "c", "d"]
+        .into_iter()
+        .map(VariableName::from)
+        .map(Into::into);
+    int_instructions
+        .chain(bool_instructions)
+        .chain(exec_instructions)
+        .chain(variables)
+        .collect()
```
JustusFluegel, line 214: No need to collect here, just return a `impl Iterator<...>` instead

packages/push/examples/smallest/main.rs
```diff
@@ -0,0 +1,215 @@
...
+    let exec_instructions = ExecInstruction::iter()
+        // The `ExecInstruction::DupBlock` instruction often leads to substantially more complicated
+        // evolved programs which take much longer to run. Restore this `filter` line
+        // to remove it from the instruction set.
+        .filter(|&i| i != ExecInstruction::dup_block())
+        .map(Into::into);
+    let variables = ["a", "b", "c", "d"]
+        .into_iter()
+        .map(VariableName::from)
+        .map(Into::into);
```
JustusFluegel, line 209: I would add newlines around this

JustusFluegel (changes requested):
I added some comments - some of this might stem from this pr being older than the recent-ish example refactors. It might be advisable to generally compare this with the other examples so that they mostly align

packages/push/examples/smallest/args.rs
NicMcPhee: I totally never looked at `args.rs`, did I? Sighs. Glad you were more attentive!

packages/push/examples/smallest/main.rs
NicMcPhee: I really like that idea. I need to get in the habit of using distributions more often since they're a really natural fit here in the EC work.

packages/push/examples/smallest/main.rs
NicMcPhee: We choose to stick with this approach (`main()` on top); there's a new issue #188 to update the other examples to match this approach.

packages/push/examples/smallest/main.rs
NicMcPhee: I chose to keep the `structs` for `Input` and `Output`, in part so that the output `i64` couldn't be confused with other kinds of `i64`s (like inputs).

JustusFluegel (approved):
If the tests pass this LGTM

PR #153, opened 2024-05-05
Enable lints

JustusFluegel:
Enables some more clippy lints


JustusFluegel:
All failing lints are documented in #167

scripts/issue_creator.nu
NicMcPhee, line 1: This was really cool! I like having both the master issue and all the sub-issues for each lint.

packages/ec-linear/examples/count_ones/main.rs
NicMcPhee, line 1: Do we want issues for the examples as well? I'm not sure about that, but I feel like we don't want to "publish" examples with things like `dbg!()` in them?

NicMcPhee (approved):
This is awesome – thanks!

PR #183, opened 2024-05-14
Rename `WithTarget` with `WithTargetFn`

NicMcPhee:
I think this is more consistent with some other naming we've done, and it was what I imagined this would be called when I was trying to find it while doing some other refactoring.


JustusFluegel (approved):
I like the general spirit of this, although I somehow find the `_fn` suffix not very rust-y. I would probably prefer something else.

But since I don't have any better idea at the moment, I'll approve this, this should arguably be an improvement from before. Might want to revisit it at some point tho.

NicMcPhee:
I'm definitely open to a better name when you think of one. 💯

PR #184, opened 2024-05-14
Derive `EnumIter` for `ExecInstruction`

NicMcPhee:
This derives `strum::EnumIter` for `ExecInstruction`. For this to work, though, I had to also derive `Default` all the "sub-types" (e.g., `When`). That arguably should be been done before. While I was there, I also derived `Copy` for those that didn't already have it.

I also switched `complex_regression/main.rs` to use the new `Display` for individuals instead of `Debug` as before.


JustusFluegel (approved):
LGTM

PR #194, opened 2024-05-19
Set the labels used explicitly for dependabot

JustusFluegel:


NicMcPhee (approved):
That's great – thanks!

Issue #191, opened 2024-05-19
Use `saturating_neg` for integer negation


NicMcPhee:
The `IntInstruction::Negate` can fail if the value is the smallest possible negative integer, because negating that doesn't have a corresponding positive value. This is a very unlikely situation, but did come up in some CLI tests, so we should use `saturating_neg()` instead.

We should also add a test that fails before fixing the problem.

PR #199, opened 2024-05-20
Use `saturating_neg()` for integer negation (issue #191)

NicMcPhee:
There's an edge case for the `Negate` instruction when the value is `i64::MIN` since negating that generates a wrapping error. This is because the two's complement representation used for integer types has an asymmetry where the magnitude of `MIN` is one larger than the magnitude of `MAX`. See <https://en.wikipedia.org/wiki/Two%27s_complement#Most_negative_number> for additional details and examples.

My solution here is to have `Negate` return `i64::MAX` when it negates `i64::MIN`. This isn't mathematically accurate, but is logically plausible since it converts the smallest negative number into the largest positive number.

The alternative would be to have this instruction "fail" in some way in this case, presumably by either being skipped or
generating a fatal error and terminating program evaluation. Neither of these options seem terribly reasonable from an evolutionary standpoint. 
- Skipping would leave the value on the stack, not negated, which would
be quite "surprising" and almost certainly leading to unexpected behavior.
- Failing doesn't seem to be in the spirit of Push, where we try to make
sure almost every instruction "succeeds" in some reasonable way.

So I went saturation. I'm open to alternatives, though.

While I was here, I pulled `IntInstruction` into a new `int` module, and pulled `Negate` out into a sub-module. I also added documentation for `Negate` in the style we've been using for `ExecInstruction`s. That makes this bigger than strictly necessary, but it seemed a reasonable refactoring while I was there.

Closes: #191


packages/push/src/instruction/int/negate.rs
```diff
@@ -0,0 +1,132 @@
...
+            .unwrap()
+            .with_no_program()
+            .build();
+        let result = IntInstruction::negate().perform(state).unwrap();
+        assert_eq!(result.stack::<i64>().size(), 1);
+        assert_eq!(*result.stack::<i64>().top().unwrap(), i64::MAX);
+    }
+
+    #[proptest]
+    fn negate(#[any] x: i64) {
```
NicMcPhee, line 118: I need to fix this `proptest` for `Negate` so that it won't fail if the value generated by `any` is `i64::MIN`.

NicMcPhee: This is now dealt with in the following commit.

NicMcPhee:
I just realized that I should add a `# Differences` section to the documentation before we merge this.

NicMcPhee:
After some contemplation and a conversation with Google Gemini, I went with an orange badge saying "Saturates on overflow":

image

I'm totally open to alternative suggestions for color and/or wording, however.

I also used Shields.io for this, although we could just use a little CSS in our Rustdocs and avoid the dependency on an outside service. Thoughts on that?

NicMcPhee:
One thing I noticed is that putting the Markdown to include the badge at the top of the documentation for `Negate` means that the alt-text for that image ends up being the "short form" documentation for `Negate` on the `push::instruction::int::negate` page:

image

This doesn't seem awesome, but I'm not sure how to best address it. I don't want to move the badge lower on the page and I don't want to change the alt-text to be a more general description. Ideas?

JustusFluegel (changes requested):
So generally this looks very reasonable to me :) But the thing with the docs & badges seems a bit unfortunate, and I'd say we should do something about it before merging this in. I don't think it's a huge issue for this to be open a few days longer, since the edgecase is currently so rare.

JustusFluegel (approved):

Issue #201, opened 2024-05-21
Add some coherent benchmarking (perhaps to CI?)


NicMcPhee:
We're discussing several possible changes that could have performance implications, and we don't have very complete or coherent Criterion benchmarking at the moment, so we don't really know when changes are (or aren't) impacting our performance.

It would be nice, then to do something about that.

At a minimum I should clean up the existing Criterium benchmarks, which are mostly from the very early days of the project and both incomplete and probably a bit of a mess.

It would also be nice to somehow integrate this into the CI pipeline, at a minimum informing us when there's a performance drop, and perhaps even (softly) blocking merges that hurt performance in a meaningful way. I've never looked into how one would do this, though. There appears to be a [Criterion Compare PRs](https://github.com/marketplace/actions/criterion-compare-prs) GitHub Action that would presumably be useful here. They highlight the fact that you don't control the run-time environment (GitHub's VMs), though, so it's very possible that there will be changes in measured performance times simply caused by changes in load, etc., on their servers.

Issue #20, opened 2023-11-18
Handle overflow in total error


NicMcPhee:
If a set of individual errors is large enough, their sum can currently overflow because `.sum()` doesn't do checked addition. Some options include

- Using a floating point type for the total error
- Using `i128` for the total error if the individual errors are `i64`
- Use a package like [apint](https://docs.rs/apint/latest/apint/) to provide arbitrary precision integers for the total error.

I kinda like the last one, although I suspect using an `i128` would also work, at least for `i64` errors. If we wanted `i128` errors, though, `apint` would be the better choice.

NicMcPhee:
I came back to this while playing with implementations of two new PushGP problems (Smallest and Median), where it was often the case that while the individual errors on individual training cases fit in a `i64`, adding up 100 of them wraps around, and I sometimes ended up with negative total errors (when all errors should be ≥0).

This is bad on several levels, including the fact that selectors that depend on total error (like `Best` and `Tournament`) then "like" these (very bad) individuals and select them enthusiastically. It doesn't affect `Lexicase` (yay!), but since I use `Best` to pick one to display at each generation that display isn't super helpful. 

As mentioned in the initial issue, the problem is that we use `.sum()` to add up all the errors in several places in [`test_results.rs`](https://github.com/unhindered-ec/unhindered-ec/blob/main/packages/ec-core/src/test_results.rs), and `.sum()` wraps. Saturating seems to be the right answer, but `.sum()` isn't supported on `Saturating` types. There have been [discussions in the Rust world about implementing `Sum` for `Saturating` types](https://github.com/rust-lang/libs-team/issues/303), but they appear to have gotten stuck in questions of how to handle short-circuiting (which probably isn't a terribly important issue for us).[^1]

[^1]: There's also an interesting question of how to handle things like `1i64 + i64::MAX + (-2i64)`. The OP on the linked thread is really only concerned about unsigned types, so this doesn't come up for them. As we've used `Score` and `Error` so far they are in fact effectively unsigned types, but I've used signed types like `i64` to hold them. This might suggest that this wasn't a great choice? Maybe there's another issue to be created here?

I'm not sure how to best handle this? At the moment I have a hack in my version of Smallest that simply checks if the total is negative and changes it to `::MAX` if it is. This isn't great, though, because (a) it would have to be done on a problem-by-problem basis and (b) it's (dimly) possible that we could wrap so far we end up positive again and a check like this wouldn't be able to tell.

[...]

PR #204, opened 2024-05-26
Benchmark `TestResult`

NicMcPhee:
This adds some simple Criterion benchmarks for the `TestResults` type. The goal is to have the ability to measure the performance of the `TestResults` implementation in anticipation of generalizing it (#20), so we can see if that generalization has any impact on the performance.

We bench two particular behaviors:

- Constructing a `TestResults` from an array of values
   - This may not be perfect because in practice we're typically constructing them from a `Vec` of values, but I'm assuming that the iterators over arrays and `Vec`s have similar performance. (Let's hope that's true!)
- Finding the smallest (by total) in a `Vec<TestResults>`

The reason for benchmarking these is that the planned generalization will make the total result have a potentially different type than individual results, and that total result type might be "bigger" (and thus possibly "slower") than the individual types. We might have `i64` for individual results, but `i128` for the total. Doing addition and comparison operations on `i128`s will presumably be slower (possibly by only a tiny amount) than working on `i64`s, so it would be good to have benchmarks so we can see if it actually matters.

This does part of #201.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
+name = "anes"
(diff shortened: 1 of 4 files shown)


JustusFluegel (approved):
LGTM

JustusFluegel (dismissed):

JustusFluegel (approved):

PR #205, opened 2024-05-28
Implement Median

NicMcPhee:
This implementation of the Median problem was mostly done by @GregoryPetterson with some help from me. I then did some clean-up at the end so it was closer to the model we'd worked out for the Smallest problem.


packages/push/examples/median/main.rs
```diff
@@ -0,0 +1,210 @@
...
+    fn median(&self) -> Output {
+        let Self(mut input) = self;
+        input.sort_unstable();
+        Output(input[1])
+    }
+}
+
+impl Distribution<Input> for Uniform<i64> {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Input {
+        Input([self.sample(rng), self.sample(rng), self.sample(rng)])
```
JustusFluegel, line 52: This is probably nicer with
```rs
std::array::from_fn(|| self.sample(rng))
```
( didn't know that existed until a while ago, should probably do the same change for smallest )

packages/push/examples/median/main.rs
```diff
@@ -0,0 +1,210 @@
...
+fn compute_error(final_state: &PushState, penalty_value: i128, expected: i64) -> i128 {
+    final_state
+        .stack::<i64>()
+        .top()
+        .map_or(penalty_value, |answer| {
+            i128::from(*answer)
+                .saturating_sub(i128::from(expected))
+                .abs()
+        })
+}
```
JustusFluegel, line 186: These are swapped in `smallest` - do we care enough to keep a consistent order?

packages/push/examples/median/main.rs
```diff
@@ -0,0 +1,210 @@
...
+
+fn instructions() -> impl Iterator<Item = PushInstruction> {
+    let int_instructions = IntInstruction::iter()
+        // Restore this line to remove `Min` from the instruction set.
+        // .filter(|&i| i != IntInstruction::Min)
```
JustusFluegel, line 191: Does this make sense in median?

packages/push/examples/smallest/main.rs
```diff
@@ -111,8 +109,6 @@ fn main() -> Result<()> {
         .into_collection_generator(population_size)
         .sample(&mut rng);
 
-    ensure!(population.is_empty().not());
```
JustusFluegel, line 114: Why was this removed?

JustusFluegel (approved):
Lgtm, added a few minor comments.

packages/push/examples/median/main.rs
NicMcPhee: Probably 😆

packages/push/examples/median/main.rs
NicMcPhee: It probably does. If you have both `min` and `max` then it's easier. If you don't have both, then you need some fancier stuff.

packages/push/examples/smallest/main.rs
NicMcPhee: I'm happy to put it back if you think it's useful. I think it was a leftover from early on when I was writing the first `main` examples, and then I just wanted to make sure that things were working. It seemed like a "leftover" to me, so I removed it.

JustusFluegel: I think this was probably a useful check, at least it seemed that way to me - I would just leave it in.

packages/push/Cargo.toml
```diff
@@ -45,6 +45,7 @@ macros = ["dep:push_macros"]
 workspace = true
 
 [[package.metadata.example_runner.examples.smallest]]
+[[package.metadata.example_runner.examples.median]]
```
JustusFluegel, line 48: doesn't this override the old config for smallest? and below as well?

JustusFluegel (approved):
LGTM - feel free to merge if you want to @NicMcPhee :)

PR #206, opened 2024-05-29
Set GitHub Actions `fail-fast` to `false`

NicMcPhee:
This ensures that all three toolchains will run even if one of them fails. Currently if `nightly` fails, then that can prevent either `stable` or `beta` from running to completion.


JustusFluegel (approved):
LGTM

PR #208, opened 2024-06-01
Fix clippy lints

JustusFluegel:
This fixes the lint config in cargo.toml as well as fixes some newly introduced nightly lints like `clippy::doc_lazy_continuation` which checks for not indented doc comments that are probably meant to be inside the list item.

This also adds a override to the helix config to always run the lints with nightly - I don't know if we want this in vscode as well.


Cargo.toml
```diff
@@ -37,12 +37,17 @@ push_macros = { path = "packages/push-macros" }
 
 
 [workspace.lints.clippy]
-pedantic = "warn"
-nursery = "warn"
-cargo = "warn"
-perf = "warn"
-complexity = "warn"
-correctness = "deny"
+# Lint groups
```
NicMcPhee, line 40: Might it be nice to have a comment that says something about why all these have a priority of -1?

Presumably the reason we need to add this is because we have all the individual lints, right? I haven't chased all those down, but I presume there's not some group that would include all of them for us?

Cargo.toml
```diff
@@ -65,7 +70,6 @@ unnecessary_safety_doc = "warn"
 unnecessary_self_imports = "warn"
 unneeded_field_pattern = "warn"
 use_debug = "warn"
-style = "warn"
```
NicMcPhee, line 68: Why did this get removed?

NicMcPhee (commented):
All this looks fine, although I did have a few questions.

Cargo.toml
JustusFluegel: Maybe some of them, but not all, no. The reason for the -1 is that that ensures that the individual lints (with a default priority of 0 ) override the groups, if they are different.

Cargo.toml
JustusFluegel: That didn't get removed, it just was moved up to the lint groups, cuz it is one.

Cargo.toml
JustusFluegel: I'll add a comment in the file explaining it.

NicMcPhee: That's actually what I intended, but didn't say very clearly. Thanks!

NicMcPhee (approved):

PR #209, opened 2024-06-01
Remove all occurances of the rust-ga name and change it to unhindered_ec

### LINKED ISSUE #200: Change `rust-ga` to `unhindered-ec` in Github Actions workflows
[The GitHub Actions workflow](https://github.com/unhindered-ec/unhindered-ec/blob/main/.github/workflows/ci.yml) refers to `rust-ga`. Given that we've changed the name of the umbrella project to `unhindered-ec`, we should probably update the workflow. It might also be good to search for any other occurrences of `rust-ga` and update them as well.

JustusFluegel:
Closes #200 

( i searched through all files with a pretty simple nushell script so i am reasonably confident this should cover everything )


JustusFluegel:
I also updated the ruleset to reflect the new names  - might make it necessary to merge this in first

NicMcPhee (commented):
This is great – thanks!

My one concern is removing all the old benchmark files. I think there's logic in some of those that might be worth preserving and moving into "appropriate places", and that's easier if they're still in the system. We can obviously go back into the history and find them, but that's gonna be more of a pain.

Thoughts?

JustusFluegel:
> This is great – thanks!
> 
> My one concern is removing all the old benchmark files. I think there's logic in some of those that might be worth preserving and moving into "appropriate places", and that's easier if they're still in the system. We can obviously go back into the history and find them, but that's gonna be more of a pain.
> 
> Thoughts?

I would say then let's go back in the history when we need them, because they aren't doing anything right now (cuz they aren't even in a package ) and are mostly dead-weight.

NicMcPhee:
I'm sold!

NicMcPhee (approved):

PR #215, opened 2024-06-23
Make `push::error::stateful::new_boxed` const

NicMcPhee:
Add `const` to `new_boxed`. This hopefully fixes a problem where `nightly` wasn't building successfully, but I'm not sure what's going to happen on `stable` and `beta` now.

I'm assuming (based on a suggestion by @JustusFluegel) that there's some change in `nightly` that now allows functions that take `Box` values to be `const`, although I can't find anything online that would explain this change.


JustusFluegel (approved):

JustusFluegel:
Don't forget your labels :)

PR #216, opened 2024-06-23
Add associated `Error` type to `Operators`

NicMcPhee:
This adds an associated `Error` type for the `Operator` trait and changes the return type for `apply` to be a `std::result::Result` instead of an `anyhow::Result`.

This forced changes in anything that implemented `Operator`, as well as several things that acted on `Operator`s.

In several cases we had to add bounds saying that `anyhow::Error: From<X::Error>` to ensure that we could convert associated errors to `anyhow::Error`s. This (and other uses of `anyhow` in `ec_core`) should ultimately be removed as we transition to `miette` errors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,8 @@ itertools = "0.12.1"
 macro_railroad_annotation = "1.0.3"
 test-strategy = "0.3.1"
 proptest = "1.2.0"
-criterion = { version = "0.5.1" }
-miette = { version = "7.2.0" }
+criterion = "0.5.1"
+miette = "7.2.0"
 
 ec-core = { path = "packages/ec-core" }
 ec-linear = { path = "packages/ec-linear" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ macro_railroad_annotation = "1.0.3"
 test-strategy = "0.3.1"
 proptest = "1.2.0"
 criterion = { version = "0.5.1", features = ["html_reports"] }
+miette = { version = "7.2.0", features = ["fancy"] }
 
 ec-core = { path = "packages/ec-core" }
 ec-linear = { path = "packages/ec-linear" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,8 +28,8 @@ itertools = "0.12.1"
 macro_railroad_annotation = "1.0.3"
 test-strategy = "0.3.1"
 proptest = "1.2.0"
-criterion = { version = "0.5.1", features = ["html_reports"] }
-miette = { version = "7.2.0", features = ["fancy"] }
+criterion = { version = "0.5.1" }
+miette = { version = "7.2.0" }
 
 ec-core = { path = "packages/ec-core" }
 ec-linear = { path = "packages/ec-linear" }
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -29,6 +29,7 @@ macro_railroad_annotation = "1.0.3"
 test-strategy = "0.3.1"
 proptest = "1.2.0"
 criterion = { version = "0.5.1", features = ["html_reports"] }
+miette = { version = "7.2.0", features = ["fancy"] }
```
JustusFluegel, line 32: probably don't want any features enabled in the root-level cargo.toml?

packages/ec-core/src/operator/composable/repeat_with.rs
```diff
@@ -23,21 +22,28 @@ impl<F, Input, const N: usize> Operator<Input> for RepeatWith<F, N>
...
+    fn apply(
+        &self,
+        input: Input,
+        rng: &mut rand::rngs::ThreadRng,
+    ) -> Result<Self::Output, Self::Error> {
+        #[allow(clippy::panic)]
+        Ok(iter::repeat_with(|| self.f.apply(input.clone(), rng))
             .take(N)
-            .try_collect::<_, Vec<<F as Operator<Input>>::Output>, anyhow::Error>()?
+            .try_collect::<_, Vec<_>, _>()?
```
JustusFluegel, line 38: Tbh I would just use 
```suggestion
            .collect::<Result<Vec<_>, _>>()?
```
here, then we don't have the additional dependency on itertools (per itertools docs is try_collect just a shorthand for exactly that) but if you prefer this we can keep it this way.

JustusFluegel (changes requested):
Apart from the feature enabled this looks good to me :)

JustusFluegel:
And of course make sure the ci passes

Cargo.toml
NicMcPhee: Ah, I wasn't sure about whether we needed to include it here so that it would be available downstream. Apparently not. I'll fix that.

packages/ec-core/src/operator/composable/repeat_with.rs
NicMcPhee: I agree with you here. No point in bring in `try_collect` from `itertools` when `.collect()` will work fine. My guess is that when this code was written over a year ago, when I didn't really understand the ability to directly collect `Result<Vec<_>, _>`. I probably thought I "needed" `try_collect` here. There's one other use of `try_collect()` (in `generation.rs`), but I do think that's more readable as is, in part because the compiler is able to fully infer the types there with `try_collect()`, where we have to specific types if we use `collect()`.

NicMcPhee:
@JustusFluegel: Thanks for adding labels for me!

Cargo.toml
```diff
@@ -28,7 +28,8 @@ itertools = "0.12.1"
 macro_railroad_annotation = "1.0.3"
 test-strategy = "0.3.1"
 proptest = "1.2.0"
-criterion = { version = "0.5.1", features = ["html_reports"] }
+criterion = { version = "0.5.1" }
+miette = { version = "7.2.0" }
```
JustusFluegel, line 32: maybe just inline the version here? ( get rid of the object )

JustusFluegel (changes requested):
Take a look at the cargo.toml again :)

JustusFluegel (approved):
LGTM

PR #223, opened 2024-07-03
Add specific `strum(to_string = …)` for Push `Push` instructions

### LINKED ISSUE #222: Fix `Display` for `Push` instructions so they show what value is being pushed
Some recent changes have made it so that when we print out a `Push` instruction it no longer shows the value being pushed. That makes programs evolved to solve examples like `complex-regression` impossible to fully interpret.

We should probably have some kind of test for this as well, so we can catch this kind of regression in the future.

Related to #90

NicMcPhee:
We need to display the values being pushed when we use a `Push(T)` instruction for `BoolInstruction`, `IntInstruction`, or `FloatInstruction`. This uses the `#[strum(to_string = "Push({0})")]` annotation to get each of these to display with a form like `Float-Push(1)`.

~~Adding these failed with the versions of `strum` and `strum-macros` that we were using at that point. I ran `cargo update` to update all the dependencies, including those for `strum` and `strum-macros`. That fixed the display, but unfortunately lead to a Clippy warning "multiple versions for dependency `heck`: 0.4.1, 0.5.0", which causes all the checks to fail on this PR.~~
Edit: Since this was laying around for so long I (@JustusFluegel ) just reset-ted to the main branch and cherry-picked the relevant commit. The heck issue also doesn't show up anymore.

Closes #222

### DIFF
--- a/packages/push/Cargo.toml
+++ b/packages/push/Cargo.toml
@@ -23,8 +23,8 @@ miette = { workspace = true }
 ec-core = { workspace = true }
 ec-linear = { workspace = true }
 
-strum = "0.26.2"
-strum_macros = "0.26.2"
+strum = "0.26.3"
+strum_macros = "0.26.4"
 embed-doc-image = "0.1.4"
 push_macros = { workspace = true, optional = true }
 collectable = "0.0.2"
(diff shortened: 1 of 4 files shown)


JustusFluegel:
@NicMcPhee did we have a look at the `BoolInstruction::` and `IntInstruction::`, etc. prefixes yet? Let's make it consistent with how the default strum impl does it for the other enum members and then we can get this finally merged in.

JustusFluegel (approved):
Fixed it up a bit let's get it merged in!

JustusFluegel:
@NicMcPhee feel free to press the merge button

PR #224, opened 2024-07-04
Add benchmark for evaluating floating point Push program

NicMcPhee:
This uses an evolved program that successfully solves a symbolic regression problem. This should provide a reasonable check for significant changes in the performance of a simple Push "function" that has no control flow.

Could be a part of addressing #201


JustusFluegel (approved):
LGTM - is the heck ignore part of this branch tho, or should that be left for the other pr?

NicMcPhee:
That's a good question. It arguably should be part of a different PR that _precedes_ this one. The other entries in `clippy.toml` all came in together in a single PR, so it would be "good practice" to do have this addition in it's own PR. Thanks for catching that.

NicMcPhee:
I removed the bit that ignored `heck` and created a new PR for that: #230

JustusFluegel (approved):
:+1:

PR #229, opened 2024-07-07
Actually use nightly in vscode for clippy lints.

JustusFluegel:
Quick fix for a blunder in #226


NicMcPhee:
I'm not sure I understand what's happening here. Will this make it so that someone using VSCode (such as myself) will end up using nightly Rust when working on this project?

JustusFluegel:
Only for the lints, which we discussed in the other pr should probably use nightly to show all of them

NicMcPhee (approved):

PR #230, opened 2024-07-09
Ignore duplicate version of `heck` crate

NicMcPhee:
This tells Clippy to not worry about the fact that we have multiple version of the `heck` crate. Hopefully we can remove this after we get updated versions of some other crates.


JustusFluegel (approved):
LGTM

PR #232, opened 2024-07-17
Rename `CollectionGenerator` to just `Generator`

### LINKED ISSUE #220: Rename `CollectionGenerator` to `collection::Generator`
This is only of only two remaining (as of July 2024) Clippy warnings about `module_name_repetitions`. I think it would be entirely reasonable to rename it to `Generator` and just make a habit of explicitly prefixing with `collection::Generator` if it's useful to know which kind of generator we're talking about.

Listed in #48

NicMcPhee:
This eliminates a "Module name repetition" warning for the name `CollectionGenerator`.

Closes #220

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -18,6 +18,7 @@
         "Recombinator",
         "recombinators",
         "rngs",
+        "rustversion",
         "Scorable",
         "scottmcm's",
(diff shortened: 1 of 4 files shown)


JustusFluegel (approved):
LGTM

PR #233, opened 2024-07-21
Convert `if-else` chain to `match` in lexicase selection

### LINKED ISSUE #231: prefer if over match in lexicase.rs?
@NicMcPhee preferred if-else in lexicase.rs but in #228 we noticed that we might want to change it to match anyways, for consistencies sake.

NicMcPhee:
Clippy recommended using `match` and `cmp` instead of an `if-else` chain in the implementation of lexicase selection. When I first wrote this, I liked the `if-else` version and stuck with it. Coming back to it, though, I do agree that the `match` version is probably nicer, so I converted over.

To help make sure I wasn't breaking things, I took this opportunity to also add some tests to lexicase selection to provide some support for any future refactorings. The first few are fairly straightforward, acting on hard-coded populations. The last is a more complex `proptest` property-based test that generates a random population with two test cases, and then creates an individual that is guaranteed to be selected and adds that to the population, confirming that lexicase selection returns that "best" individual.

Closes #231


packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,181 @@ where
...
+    use prop::collection;
+    use proptest::prelude::*;
+
+    use super::*;
+    use crate::individual::ec::EcIndividual;
+
+    #[test]
+    fn single_best_single_error() {
+        // 9 is the sole winner
+        let population = vec![5, 8, 9, 6, 3, 2, 0]
```
JustusFluegel, line 123: don't use `vec![]` here - it's not needed, just use an array.

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,181 @@ where
...
+        let lexicase = Lexicase::new(1);
+        let mut rng = rand::thread_rng();
+
+        assert_eq!(&2, lexicase.select(&population, &mut rng).unwrap().genome());
+    }
+
+    #[test]
+    fn multiple_copies_best() {
+        // The two 9s are the possible winners
+        let population = vec![5, 8, 9, 6, 3, 2, 0, 9]
```
JustusFluegel, line 138: same here

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,181 @@ where
...
+        assert!(
+            selected == 2usize || selected == 7usize,
+            "genome {selected} should have been 2 or 7"
+        );
+    }
+
+    #[test]
+    fn single_best_multiple_errors() {
+        // [9, 8] is the sole winner
+        let population = vec![[5, 3], [8, 2], [9, 8], [6, 2], [3, 8], [2, 8], [0, 6]]
```
JustusFluegel, line 157: and here

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,181 @@ where
...
+            .into_iter()
+            .enumerate()
+            .map(|(id, es)| EcIndividual::<usize, TestResults<i32>>::new(id, TestResults::from(es)))
```
JustusFluegel, line 160: I would probably do this instead, might be more readable:
```suggestion
            .map(TestResults::<i32>::from)
            .enumerate()
            .map(EcIndividual::new)
```

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,181 @@ where
...
+
+        let lexicase = Lexicase::new(2);
+        let mut rng = rand::thread_rng();
+
+        assert_eq!(&2, lexicase.select(&population, &mut rng).unwrap().genome());
+    }
+
+    #[test]
+    fn multiple_best_multiple_errors() {
+        let population = vec![
```
JustusFluegel, line 171: here as well

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,181 @@ where
...
+            [9, 8], // A possible winner
+            [6, 2],
+            [3, 8],
+            [2, 8],
+            [0, 6],
+            [7, 9], // A possible winner
+        ]
+        .into_iter()
+        .enumerate()
+        .map(|(id, es)| EcIndividual::<usize, TestResults<i32>>::new(id, TestResults::from(es)))
```
JustusFluegel, line 183: same as above
```suggestion
            .map(TestResults::<i32>::from)
            .enumerate()
            .map(EcIndividual::new)
```

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,181 @@ where
...
+    // `EcIndividual`s (i.e., a population). I then determine the largest of the
+    // first values (i.e., the highest score on the first test case), and then the
+    // largest of the second values. I make a new set of results that is those
+    // highest values with one (randomly chosen by proptest) incremented by 1.
+    // This new result is guaranteed to be "better" than any other in the list,
+    // so it should be what is selected after I add it to the population.
+    //
+    // There are several helper functions that exist just to support the somewhat
+    // complex logic of this test.
+    #[test_strategy::proptest]
```
JustusFluegel, line 211: everywhere else we just import `test_strategy::proptest` - do we want to do that here as well?

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,181 @@ where
...
+        assert_eq!(&2, lexicase.select(&population, &mut rng).unwrap().genome());
+    }
+
+    #[test]
+    fn multiple_copies_best() {
+        // The two 9s are the possible winners
+        let population = vec![5, 8, 9, 6, 3, 2, 0, 9]
+            .into_iter()
+            .enumerate()
+            .map(|(id, e)| EcIndividual::<usize, TestResults<i32>>::new(id, TestResults::from([e])))
```
JustusFluegel, line 141: Same as below here:
```suggestion
            .map(|v| [v])
            .map(TestResults::<i32>::from)
            .enumerate()
            .map(EcIndividual::new)
```

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,181 @@ where
...
+    use super::*;
+    use crate::individual::ec::EcIndividual;
+
+    #[test]
+    fn single_best_single_error() {
+        // 9 is the sole winner
+        let population = vec![5, 8, 9, 6, 3, 2, 0]
+            .into_iter()
+            .enumerate()
+            .map(|(id, e)| EcIndividual::<usize, TestResults<i32>>::new(id, TestResults::from([e])))
```
JustusFluegel, line 126: Same as below here as well
```suggestion
            .map(|v| [v])
            .map(TestResults::<i32>::from)
            .enumerate()
            .map(EcIndividual::new)
```

JustusFluegel (changes requested):

packages/ec-core/src/operator/selector/lexicase.rs
NicMcPhee: Thanks for catching that. I started with an array, and for some reason decided I needed to change to `vec!`. It works fine as an array, though, so I don't know what I was thinking.

packages/ec-core/src/operator/selector/lexicase.rs
NicMcPhee: Adding the `From<(G, R)>` implementation was pretty straightforward. What I ended up with, though, was:

```rust
        let population = [5, 8, 9, 6, 3, 2, 0]
            .into_iter()
            .map(once)
            .map(TestResults::<i32>::from)
            .enumerate()
            .map(EcIndividual::from)
            .collect::<Vec<_>>();
```

and I'm not entirely sure which approach I prefer now. This avoids any closures, which is nice, but it's not super obvious why `.map(once)` (or `.map(|v| [v])` if you prefer) is there. Maybe I should write some helper functions that encapsulate (and document) this logic rather than repeating it across multiple tests?

packages/ec-core/src/operator/selector/lexicase.rs
NicMcPhee: Good catch. I imported `proptest::prelude::*` (which is what they do in the `proptest` book), and that created a conflict that I dealt with by qualifying `test_strategy::proptest` here.

I've replaced the glob import with several specific imports, which avoids that conflict and we can just use `#[proptest]` here again.

packages/push/src/push_vm/stack.rs
```diff
@@ -261,10 +261,13 @@ impl<T> Stack<T> {
...
-            num_requested: 2,
-            num_present: self.size(),
-        })?;
+        let index_second_to_top =
+            self.size()
+                .checked_sub(2)
+                .ok_or_else(|| StackError::Underflow {
+                    num_requested: 2,
+                    num_present: self.size(),
+                })?;
```
NicMcPhee, line 270: This change was made to address a new Clippy warning in nightly. Since there's a function call (`self.size()`) embedded in the expression, the warning recommends using `.ok_or_else()` to make the closure lazy, and potentially avoid that function call when it isn't necessary.

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,200 @@ where
...
+
+    use super::*;
+    use crate::individual::ec::EcIndividual;
+
+    // Generate a population from a vector of single scores, one per individual.
+    // These have to be converted to iterators using `once` so they're "like"
+    // arrays of scores.
+    fn population_from_single_scores<I>(scores: I) -> Vec<EcIndividual<usize, TestResults<i32>>>
+    where
+        I: IntoIterator<Item = i32>,
```
JustusFluegel, line 131:
```suggestion
    fn population_from_single_scores(scores: impl IntoIterator<Item = i32>) -> Vec<EcIndividual<usize, TestResults<i32>>>
```

this could be simplified like so :)

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,200 @@ where
...
+        // an iterator over `i32`s.
+        population_from_scores(scores.into_iter().map(once))
+    }
+
+    // Generate a population from a collection of score collections (e.g., arrays),
+    // one per individual.
+    fn population_from_scores<I, J>(scores: I) -> Vec<EcIndividual<usize, TestResults<i32>>>
+    where
+        I: IntoIterator<Item = J>,
+        J: IntoIterator<Item = i32>,
```
JustusFluegel, line 143: And here as well:

```suggestion
    fn population_from_scores<I, J>(scores: impl IntoIterator<Item: IntoIterator<Item = i32>>) -> Vec<EcIndividual<usize, TestResults<i32>>>
```

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,200 @@ where
...
+    fn multiple_copies_best() {
+        // The two 9s are the possible winners
+        let population = population_from_single_scores([5, 8, 9, 6, 3, 2, 0, 9]);
+
+        let lexicase = Lexicase::new(1);
+        let mut rng = rand::thread_rng();
+
+        let selected = *lexicase.select(&population, &mut rng).unwrap().genome();
+        assert!(
+            selected == 2usize || selected == 7usize,
```
JustusFluegel, line 175:
```suggestion
            [2usize,7usize].contains(selected)
```

this is probably easier to maintain - if you prefer the way you've done it here this is totally fine by me as well tho.

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,200 @@ where
...
+            [0, 6],
+            [7, 9], // A possible winner
+        ]);
+
+        let lexicase = Lexicase::new(2);
+        let mut rng = rand::thread_rng();
+
+        let selected = *lexicase.select(&population, &mut rng).unwrap().genome();
+        assert!(
+            selected == 2usize || selected == 7usize,
```
JustusFluegel, line 210: Same here :)

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,200 @@ where
...
+        #[strategy(make_pop(#pop_size))] mut population: Vec<TestIndividual>,
+        #[strategy(0..1usize)] score_to_increase: usize,
+    ) {
+        prop_assert!(pop_size > 0);
+        prop_assert_eq!(pop_size, population.len());
+
+        let (winning_label, winning_individual) =
+            make_winning_individual(&population, score_to_increase);
+
+        population.push(winning_individual);
```
JustusFluegel, line 242: We could probably move this into a strategy since it's kind of a input but this is totally fine as well :+1:

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -95,3 +100,200 @@ where
...
+            EcIndividual::new(winning_label.clone(), TestResults::from(winning_scores));
+        (winning_label, winning_individual)
+    }
+
+    // Get the largest test case value in the given population from the test case
+    // specified by `index`.
+    fn largest_test_case_value(population: &[TestIndividual], index: usize) -> u16 {
+        population
+            .iter()
+            .map(|i| i.test_results.results[index])
```
JustusFluegel, line 295: I see what you did here - using `[...]` indexing :joy:  ( but since it's in a test it's fine )

packages/push/Cargo.toml
```diff
@@ -32,12 +31,13 @@ collectable = "0.0.2"
...
 easy-cast = "0.5.2"
 macro_railroad_annotation = { workspace = true }
-test-strategy = { workspace = true }
 criterion = { workspace = true, features = ["html_reports"] }
 rustversion = { workspace = true }
 
 [dev-dependencies]
 clap = { version = "4.5.1", features = ["derive"] }
+proptest = { workspace = true }
+test-strategy = { workspace = true }
```
JustusFluegel, line 40: Good catch!

packages/push/src/push_vm/stack.rs
JustusFluegel: Seems Good :+1:

JustusFluegel (approved):
LGTM; had some minor suggestions but you can decide if you want to apply them, probably more of a code style question :+1:

packages/ec-core/src/operator/selector/lexicase.rs
NicMcPhee: I'll leave this for now, but the new issue #240 reminds us to come back to this.

JustusFluegel (approved):

PR #235, opened 2024-07-24
Sporadic bug: "Attempt to negate with overflow" because of `i64::abs`

NicMcPhee:
In [one of the CI runs](https://github.com/unhindered-ec/unhindered-ec/actions/runs/9800496706/job/27062439279?pr=224) for #224, the `beta` check failed with the following message:

```
[push smallest <big_population_v2>] thread 'main' panicked at /rustc/64a1fe67112931359c7c9a222f08fd206255c2b5/library/core/src/num/mod.rs:359:5:
[push smallest <big_population_v2>] attempt to negate with overflow
```

The problem is the use of `i64::abs`, which fails with this error when taking the absolute value of `i64::MIN`. (The same thing would happen for any fixed-size signed integer type.) This is essentially the same problem as in #191, but a little indirect since it's `i64::abs` that's negating the value.

While I was fixing this, I also updated some comments in `Negate`. I also converted both this and `Negate` to use the range `((i64::MIN+1)..=i64::MAX)` instead of using `filter` since `filter` can interact badly with `proptest` simplification and complification.

### DIFF
--- a/packages/push/src/instruction/int/abs.rs
+++ b/packages/push/src/instruction/int/abs.rs
@@ -83,7 +83,10 @@ where
 
     fn perform(&self, state: S) -> InstructionResult<S, Self::Error> {
         let int_stack = state.stack::<i64>();
-        int_stack.top().map(|x| x.abs()).replace_on(1, state)
+        int_stack
+            .top()
+            .map(|x| x.saturating_abs())
+            .replace_on(1, state)
     }
 }
 
--- a/packages/push/src/instruction/int/abs.rs
+++ b/packages/push/src/instruction/int/abs.rs
@@ -1,0 +1,147 @@
+use crate::{
+    error::InstructionResult,
+    instruction::{instruction_error::PushInstructionError, Instruction},
+    push_vm::{stack::PushOnto, HasStack},
+};
+
+/// An instruction that takes the absolute value of the top
+/// value on the `i64` stack.
+///
+/// There's an edge case when the value is `i64::MIN` since
+/// negating that generates a wrapping error. This is because the two's
+/// complement representation used for integer types has an asymmetry
+/// where the magnitude of `MIN` is one larger than the magnitude of `MAX`.
+/// See <https://en.wikipedia.org/wiki/Two%27s_complement#Most_negative_number>
+/// for additional details and examples.
+///
+/// In our implementation, `Abs` returns `i64::MAX` when taking the absolute
+/// value of
+/// `i64::MIN`. This isn't mathematically accurate, but is semantically
+/// plausible since it converts the smallest negative number into the largest
+/// positive number.
+///
+/// The alternative would be to have this instruction "fail"
+/// in some way on `i64::MIN`, presumably by either skipping the instruction or
+/// generating a fatal error and terminating program evaluation. Neither of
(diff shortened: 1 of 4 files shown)


packages/push/src/instruction/int/negate.rs
```diff
@@ -130,7 +133,7 @@ mod tests {
     // We need to make sure that `x` is greater than `i64::MIN` since
     // we handle that case differently. This is described in the documentation
     // for `Negate`, and handled in the preceding test.
-    fn negate(#[filter(#x > i64::MIN)] x: i64) {
+    fn negate(#[strategy(((i64::MIN+1)..=i64::MAX))] x: i64) {
```
JustusFluegel, line 136:
```suggestion
    fn negate(#[strategy((i64::MIN+1)..=i64::MAX)] x: i64) {
```

packages/push/src/instruction/int/abs.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[test]
+    fn abs_with_i64_min() {
+        let input = i64::MIN;
+        let state = PushState::builder()
+            .with_max_stack_size(1)
+            .with_int_values(std::iter::once(input))
+            .unwrap()
+            .with_no_program()
+            .build();
+        let result = IntInstruction::abs().perform(state).unwrap();
```
JustusFluegel, line 127: Please make sure to not rely on the "instruction group" for tests for a particular instruction:
```suggestion
        let result = Abs.perform(state).unwrap();
```

packages/push/src/instruction/int/abs.rs
```diff
@@ -0,0 +1,150 @@
...
+    // we handle that case differently. This is described in the documentation
+    // for `Abs`, and handled in the preceding test.
+    fn abs(#[strategy((i64::MIN+1)..=i64::MAX)] x: i64) {
+        let state = PushState::builder()
+            .with_max_stack_size(1)
+            .with_int_values(std::iter::once(x))
+            .unwrap()
+            .with_no_program()
+            .build();
+        let result = IntInstruction::abs().perform(state).unwrap();
```
JustusFluegel, line 143: Same here:
```suggestion
        let result = Abs.perform(state).unwrap();
```

packages/push/src/instruction/int/negate.rs
JustusFluegel: And probably for this test as well (the `IntInstruction::foo()` -> `Foo` thing), but we might want to do it in another pr since this isn't actually related to `Abs`

JustusFluegel (changes requested):
Mostly lgtm, added 2 minor change requests :)

packages/push/src/instruction/int/negate.rs
NicMcPhee: Actually, I'm not entirely sure what you mean by `IntInstruction::foo()`, etc., here.

packages/push/src/instruction/int/abs.rs
NicMcPhee: Good catch – thanks. We might want to fix that on `Negate` since we're already changing things there.

JustusFluegel (approved):
LGTM

Issue #236, opened 2024-07-24
Add additional helper constructors for other common tournament sizes


NicMcPhee:
I added a `Tournament::binary()` to create an instance of binary tournament selection, encapsulating the need to convert a `usize` to `NonZeroUsize` and the necessary `.unwrap()`. It would potentially be useful to also create helpers for other common tournaments, although it's less clear what exactly those would be.

JustusFluegel:
@NicMcPhee is this still relevant with the `::of_size<N>()` const constructor?

PR #237, opened 2024-07-24
Ensure that tournament sizes are positive

### LINKED ISSUE #214: Ensure that tournament sizes are positive
Currently we allow 0 as a tournament size in `Tournament::new()`, and that actually doesn't make any sense. We should use `NonZeroUsize` for the tournament size.

NicMcPhee:
Convert `Tournament.size` to be `NonZeroUsize`, since we don't ever want size zero tournaments.

This  has the unfortunate side-effect of now requiring an `.unwrap()` or similar whenever constructing a `Tournament` since we have to convert a `usize` to a `NonZeroUsize`, which could fail. I added a `binary()` constructor that creates a binary tournament selector and handles the `unwrap()`, simplifying that fairly common use case. We could add additional tournament constructors for other common tournament sizes (see #236).

I had to update several examples to use the new `Tournament::binary()`.

As part of this I added some simple tests for tournament selection. We check that

- Selecting using a tournament size of one always returns a member of the population,
- Selecting using a tournament size of two from a population with size two always returns the larger value,
- Selecting using a tournament of size 2 from a population of size 3 containing distinct values always returns a value greater than the minimum value.

Closes #214


packages/ec-core/src/operator/selector/tournament.rs
```diff
@@ -1,18 +1,31 @@
...
+    pub const fn new(size: NonZeroUsize) -> Self {
         Self { size }
     }
+
+    /// Construct a binary tournament selector, i.e., a tournament
+    /// selector that selects two random individuals from the population
+    /// and returns the "better" of the two.
+    #[must_use]
+    pub const fn binary() -> Self {
+        Self::new(NonZeroUsize::MIN.saturating_add(1))
```
NicMcPhee, line 27: This rather odd way of creating `2` allows `binary()` to be `const`, and avoids any chance of failure and need for `unwrap()` calls. That then avoids the need for all the complex `allow/expect` stuff here.

packages/ec-core/src/operator/selector/tournament.rs
```diff
@@ -26,15 +39,91 @@ where
...
+    }
+
+    #[proptest]
+    fn tournament_size_2_pop_size_3(
+        #[strategy(-1000..1000)] x: i32,
+        #[strategy(-1000..1000)] y: i32,
+        #[strategy(-1000..1000)] z: i32,
+    ) {
+        let mut rng = thread_rng();
+        let scores = &[3 * x, 3 * y + 1, 3 * z + 2];
```
NicMcPhee, line 118: This just ensures that all three values are distinct, which allows me to use use `>` (instead of `>=`) in the assertion below.

packages/ec-core/Cargo.toml
```diff
@@ -33,6 +33,8 @@ rand = { workspace = true, features = ["alloc", "small_rng"] }
 criterion = { workspace = true }
 miette = { workspace = true, features = ["fancy"] }
 thiserror = { workspace = true }
+proptest.workspace = true
+test-strategy = "0.4.0"
```
JustusFluegel, line 37: don't we have test-strategy in the workspace yet? If we do not we should move it there and change all occurrences in dependencies to use that ( which I expect to be exactly one in that case )

packages/ec-core/src/operator/selector/tournament.rs
```diff
@@ -1,18 +1,31 @@
...
 impl Tournament {
+    /// Construct a tournament selector with the given size. This
+    /// will select `size` individuals from the population, randomly
+    /// without replacement, and return the "best" from that set.
     #[must_use]
-    pub const fn new(size: usize) -> Self {
+    pub const fn new(size: NonZeroUsize) -> Self {
         Self { size }
     }
+
```
JustusFluegel, line 21: How about we add a `::of_size<N>()` constructor like so:
```suggestion
#[must_use]
pub const fn  of_size<const N: usize>() -> Self {
   Self::new(const { match NonZeroUsize::new(N) {
       Some(x) => x,
       None => panic!("only positive tournament sizes are permitted");
   } }))    
}
```

This would enable us to drop `binary()` if we wish to and construct arbitrary sizes of tournaments with const code, but not from dynamic values like variables. Since the `panic!()` is in a `const` block it only gets executed at compile time and as such all errors that could happen will occur at compile time.

packages/ec-core/src/operator/selector/tournament.rs
JustusFluegel: just do `tournament_size_2_pop_size_3(#[any] a: i32, #[filter(|x| x != #a)] b: i32, #[filter(|x| x != a && x != b)] c: i32)` - yes `filter` isn't the best thing ever but since this case is so rare it shouldn't really matter in practice.

packages/ec-linear/examples/hiff/main.rs
```diff
@@ -1,10 +1,10 @@
 // TODO: since inner attributes are unstable, we can't use rustversion here.
 // Once we revert this commit, this is proper again.
+#![allow(clippy::allow_attributes_without_reason)]
 #![allow(
-    clippy::allow_attributes_without_reason,
     clippy::arithmetic_side_effects,
-    // reason = "The tradeoff safety <> ease of writing arguably lies on the ease of writing side \
-    //           for example code."
+//     reason = "The tradeoff safety <> ease of writing arguably lies on the ease of writing side \
+//               for example code."
```
JustusFluegel, line 7: Can we please not change this (yet) since it will make it way harder to revert everything once 1.81 lands - yes I know I will have to do a bunch of manual cleanup with allows() that were added later but I'd rather not make it more bothersome than it is already.

JustusFluegel (changes requested):
I added a few comments :)

packages/ec-linear/examples/hiff/main.rs
JustusFluegel: And just for your understanding since you seemed not very clear about that on stream: this **will** come out with 1.81, which is the current nightly version - it was stabilized already (is not a #[feature()] gate anymore), in that version. The thing why it currently only works in nightly is that  nightly is on 1.81 already, whereas beta is on 1.80 and stable on 1.79 ( well till tomorrow, tomorrow nightly will move to 1.90, beta to 1.81 and stable to 1.80 ) - the release cycle is just that everything lands in nightly first and then moves through beta and to stable.

The only case where something gets `de-stabilized` (which is actually usually a pretty huge deal and happens almost never) is when a deal-breaking bug gets found during nightly or beta testing.

packages/ec-core/Cargo.toml
NicMcPhee: Duh – good catch. `test-strategy` is definitely already in the top level `Cargo.toml`, and I should use `workspace = true` here.

packages/ec-core/src/operator/selector/tournament.rs
NicMcPhee: I could go that route, but my sense from reading the `proptest` book in various airports (and with varying amounts of sleep deprivation!) is that it really works "with" `proptest` if you avoid `filter` and that you're better off generating values with the properties you desire.

Since I'm not testing anything about the specific values of the three values (all I care about is their relative ordering), I don't see any particular downside to generating three values that are guaranteed to be distinct, and this seemed a straightforward way of doing that. I've now added a comment to this line explaining what I'm doing here, though, since it's not super obvious.

packages/ec-linear/examples/hiff/main.rs
NicMcPhee: Thanks – I didn't really understand that terribly well.

At some point we had to change some of these because I hadn't invented `Tournament::binary()` yet. I may not have reverted everything cleanly after I brought that in, however – sorry. I'll try to clean that up.

packages/ec-core/src/operator/selector/tournament.rs
```diff
@@ -1,18 +1,60 @@
...
+    ///
+    /// # Examples
+    ///
+    /// Create a tournament selector with tournament size 3:
+    /// ```
+    /// # use ec_core::operator::selector::tournament::Tournament;
+    /// let selector = Tournament::of_size::<3>();
+    /// ```
+    ///
+    /// # Panics
```
JustusFluegel, line 36: Actually this is kind of misleading since this can *never* panic at runtime, only during compile-time, which will then show a error during compilation - maybe the clippy `clippy::missing_panics_doc` lint doesn't play super nice here? I would definitely at least note that here that this is not runtime behaviour

packages/ec-core/src/operator/selector/tournament.rs
JustusFluegel: If you want to go this route I'd create a strategy for that, so that we can write something like `#[strategy(uniqe_nums(3 /* count */, 0 /* index */, -1000..1000))]` and that we needn't do that in the actual test body.

JustusFluegel: or even more straightforward:

```
fn foo(#[strategy(unique_nums())] [a,b,c]: [i32,3]) { ... }
```

NicMcPhee: That's a nice idea. Does it need to be part of this PR, or can/should that be part of a different issue/PR?

JustusFluegel: I like the idea of it beeing part of this pr but I'll leave it up to you to decide.

JustusFluegel (dismissed):

JustusFluegel (approved):

Issue #245, opened 2024-07-31
Split `Case` and `Cases` into separate files, with some module flattening


NicMcPhee:
Having `Case` and `Cases` in the same file is getting quite long, especially after we move the tests in `tests/cases.rs` into the main module. I think we want to split these into separate files to help keep things under control, but then re-export them from a parent module in a way that reduces the depth of the import statements.

- [x] Split `Case` and `Cases` into separate modules.
- [x] Split `Cases` up into sub-modules as a way of making all that code more manageable.
- [x] Move tests from `tests/cases.rs` into the appropriate new module.
- [x] Add some exports to "flatten" all this so importing isn't too tedious.

It's possible that we actually want to move `Case` and `Cases` into a different package than `push` since they're arguably more general concepts than for just Push-based genetic programming. Maybe a new `software-synthesis` package?

Should wait until #246 is done before moving on to this.

Tracked by #84

PR #247, opened 2024-07-31
Improve docs and tests for cases

### LINKED ISSUE #246: Add documentation and doctests for `Case` and `Cases`
There's currently no documentation or testing for these types. I think we can use doctest for most of the simple tests, and then maybe use `proptest` for some fancier tests if we feel that's useful/necessary.

Tracked by #84

NicMcPhee:
This adds comments and doctests for all the methods in `Case` and `Cases` and their various trait implementations.

I also re-ordered the various functions in `Cases`, trying to group them by:

   * Things that create sets of `Cases`
   * Properties of `Cases` (like `len()`)
   * Iterators over `Cases`

I found the tests to be a bit cleaner with the addition of `.inputs()` and `outputs()`, which return iterators over the inputs and outputs for a set of `Cases`.

I removed two TODO comments as we've created issues for those.

Closes #246

### DIFF
--- a/packages/push/src/evaluation/cases.rs
+++ b/packages/push/src/evaluation/cases.rs
@@ -1,15 +1,55 @@
+/// Simple pair of input and (expected) output used
+/// for testing evolved systems.
+///
+/// See also [`Cases`].
+///
+/// # Examples
+///
+/// ```
+/// # use push::evaluation::cases::Case;
+/// #
+/// let case = Case::new("Hello", 5);
+///
+/// assert_eq!(case.input, "Hello");
+/// assert_eq!(case.output, 5);
+/// ```
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub struct Case<Input, Output = Input> {
     pub input: Input,
     pub output: Output,
 }
 
+/// Create a `Case` from an input-output pair
+///
+/// # Examples
+///
+/// ```
+/// # use push::evaluation::cases::Case;
+/// #
+/// let case = Case::from(("this", 4));
+///
+/// assert_eq!(case.input, "this");
+/// assert_eq!(case.output, 4);
+/// ```
 impl<Input, Output> From<(Input, Output)> for Case<Input, Output> {
     fn from((input, output): (Input, Output)) -> Self {
         Self { input, output }
     }
 }
 
+/// Convert a `Case` into an input-output pair.
+///
+/// # Examples
+///
+/// ```
+/// # use push::evaluation::cases::Case;
+/// #
+/// let case = Case::new(true, 17);
+/// let (x, y): (bool, i32) = case.into();
+///
+/// assert_eq!(x, true);
+/// assert_eq!(y, 17);
+/// ```
 impl<Input, Output> From<Case<Input, Output>> for (Input, Output) {
     fn from(case: Case<Input, Output>) -> (Input, Output) {
         (case.input, case.output)
@@ -17,6 +57,7 @@ impl<Input, Output> From<Case<Input, Output>> for (Input, Output) {
 }
 
 impl<Input, Output> Case<Input, Output> {
+    /// Construct new `Case` from input and output.
     pub const fn new(input: Input, output: Output) -> Self {
         Self { input, output }
     }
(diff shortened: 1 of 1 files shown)


packages/push/src/evaluation/cases.rs
```diff
@@ -88,35 +236,176 @@ impl<T, Input> WithTargetFn<Input> for T
...
+    /// # use push::evaluation::cases::Cases;
+    /// #
+    /// let inputs = ["this", "and", "those"];
+    /// let cases = Cases::from_inputs(inputs, |s| s.len());
+    ///
+    /// assert!(cases.outputs().eq(&[4, 3, 5]));
+    /// ```
+    pub fn outputs(&self) -> impl Iterator<Item = &Output> {
+        self.cases.iter().map(|c| &c.output)
     }
```
JustusFluegel, line 386: if we have `.inputs()` and `.outputs()` shouldn't we have `.inputs_mut()` and `.outputs_mut()` as well as `.into_inputs()` and `.into_outputs()` as well?

JustusFluegel (dismissed):
LGTM. Added a comment, we can do that as another issue / pr if you prefer that. Do we really want the 
```
// -----------------
// .....
// -----------------
```
sections? We aren't using them anywhere else in our project and as such it seems a bit odd code-style wise. If you have such sections, I'd rather create a structure like `cases/{creation, properties, iteration}.rs` files (as a private module of course) instead of doing it with comments like you did here. But if you really prefer the comments, i don't see it as a real blocker to getting this merged.

NicMcPhee:
We definitely don't have marker comments in other places, but I suspect that's because I hadn't thought of it yet. When writing on-stream I find I rarely see the forest, focusing instead of the tree we're immediately working on. Here I stepped back and looked at the (fairly long) file, and wanted to try to organize it in some fashion, and creation, iteration, and operators seemed a plausible way. The comments seemed a way to indicate, both to myself and other readers, that this is how I'd organized things.

That said, your idea of `cases/{creation, properties, iteration}` with appropriate uses of private and export might make more sense. Is that approach common in larger Rust projects? I assume we wouldn't want to do this everywhere, but it might make sense for the larger, more complex types. (I wouldn't have necessarily thought of `Cases` as being "larger, more complex", but once we've added all the documentation and tests it does get pretty big.)

If we want to go this route, I'd be inclined to go ahead and do it as part of this PR if you're OK with that. Adding `.inputs_mut()` and `.outputs_mut()` will just make it longer, and I realized that I totally forgot to copy over the tests that are in `tests/cases.rs`, which will make it all longer still.

If we're going to break things into modules, then we might as well split up `Case` and `Cases` at the same time so we deal with all of it at once.

packages/push/src/evaluation/cases.rs
NicMcPhee: I like the idea of `.*_mut()`, but do we want/need `.into_*()`? I'm not sure when we'd want to convert a set of `Cases` into their inputs or outputs, losing/consuming the original set of cases. Is this one of those things we should do anyway to be "completist"?

NicMcPhee:
I went ahead and added `inputs_mut()` and `outputs_mut()`, which were pretty straightforward.

JustusFluegel:
Sure, fair enough, let's leave it as one file in this pr then. Do we still want `.into_inputs()` and `.into_outputs()` as well or nah?

NicMcPhee:
I'm not sure about the `into_*()` methods. I'm not sure when someone would want to convert a set of cases into just the inputs or outputs, losing the full set of cases. I guess it's possible – just because I can't imagine it doesn't mean someone won't want to do it. Do you have any thoughts on why we'd include them other than just being "complete"?

JustusFluegel:
Well we have a `.with_target_fn()` so if one wants to get their inputs back a `.into_inputs()` would be nice as kind of the `opposite` to that. Also it's pretty nice to be complete and consistent with the .iter() .iter_mut() and .into_iter() methods

JustusFluegel:
Oh and you commented about that in the comment above 2 hours ago, totally missed that, sorry.

NicMcPhee:
I just added the `.into_*()` methods. If it looks OK to you, why don't we merge this in, and do the module splitting up in the same PR where we split `Case` and `Cases` into two modules (i.e., #245)?

JustusFluegel (approved):
Sounds good. Let's get this merged in!

PR #248, opened 2024-08-04
Split `cases.rs` into separate file and modules, flattening the imports

NicMcPhee:
This splits `cases.rs` into several sub-modules:

- `case.rs` for the `Case` type
- `with_target_fn` for the `WithTargetFn` trait and implementation for `Cases`
- `iter` for the various iterator implementations for `Cases`

We used `pub use` from `cases/mod.rs` to re-export several types to flatten the import hierarchy a bit.

We also added documentation, doctests, and unit tests for all of these while we were here.

Closes: #245


packages/push/src/evaluation/cases/case.rs
```diff
@@ -0,0 +1,87 @@
...
+/// assert_eq!(x, true);
+/// assert_eq!(y, 17);
+/// ```
+impl<Input, Output> From<Case<Input, Output>> for (Input, Output) {
+    fn from(case: Case<Input, Output>) -> (Input, Output) {
+        (case.input, case.output)
+    }
+}
+
+impl<Input, Output> Case<Input, Output> {
```
JustusFluegel, line 59: do we want to add the same docs as above for the other from impl?

packages/push/src/evaluation/cases/case.rs
```diff
@@ -0,0 +1,87 @@
+/// Simple pair of input and (expected) output used
```
JustusFluegel, line 1: Some module level docs here as well? `//!`

packages/push/src/evaluation/cases/with_target_fn.rs
```diff
@@ -0,0 +1,75 @@
+use super::super::Cases;
```
JustusFluegel, line 1: do we want some module-level docs here as well?

packages/push/src/evaluation/cases/mod.rs
```diff
@@ -1,67 +1,9 @@
-/// Simple pair of input and (expected) output used
```
JustusFluegel, line 1: and module-level docs here? Idk your choice

packages/push/src/evaluation/cases/with_target_fn.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///
+    /// `T` (the type of `self`) must implement `IntoIterator<Item =
+    /// Input>`; this provides the mechanism for iterating over the
+    /// inputs so we can create the associated outputs and thus the
+    /// individual `Cases`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use push::evaluation::{Cases, WithTargetFn};
```
JustusFluegel, line 40: do we want a blank line here?

JustusFluegel (approved):
mostly lgtm, had a few questions about docs

packages/push/src/evaluation/cases/iter.rs
```diff
@@ -0,0 +1,180 @@
...
+    fn ref_into_iterator() {
+        let cases = Cases::new()
+            .with_case(Case::new("Hello", 5))
+            .with_case(Case::new("People", 6));
+
+        let mut iter = (&cases).into_iter();
+
+        assert_eq!(iter.next(), Some(&Case::new("Hello", 5)));
+        assert_eq!(iter.next(), Some(&Case::new("People", 6)));
+        assert_eq!(iter.next(), None);
```
JustusFluegel, line 139:
```suggestion
        assert!(iter.eq([&Case::new("Hello",5), &Case::new("People", 6)]);
```

JustusFluegel (dismissed):
This is fine by me as it is now, but take a look at my suggestions and see if you want to apply any of them.

packages/push/src/evaluation/cases/iter.rs
```diff
@@ -0,0 +1,180 @@
...
+        let cases = Cases::from_iter(items);
+
+        assert!(cases.into_iter().eq(items));
+    }
+
+    #[test]
+    fn ref_into_iterator() {
+        let cases = Cases::new()
+            .with_case(Case::new("Hello", 5))
+            .with_case(Case::new("People", 6));
```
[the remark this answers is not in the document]
NicMcPhee: I'm not super consistent in how I build up a list of cases in tests. Sometimes I do this (using `.with_case`). Sometimes I use `.into_iter()` like you suggested. Sometimes I use `.with_target_fn`. Etc., etc.

I think this is the most explicit and uses the least "magic" (like the implicit `.from()` call that converts pairs to `Case` in your version), so I think I lean towards it.

packages/push/src/evaluation/cases/iter.rs
NicMcPhee: Totally! And there are more of these as well. Grrr… I'll make that change and let you know when it's time for a review.

packages/push/src/evaluation/cases/iter.rs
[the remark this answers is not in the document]
JustusFluegel: Fair enough, sounds good to me

JustusFluegel (approved):
LGTM, nice testing :)

PR #253, opened 2024-08-07
Bump rand to .alpha.2 and fix renamed items

JustusFluegel:


NicMcPhee (approved):
Thanks for grinding through those for us!

Issue #244, opened 2024-07-29
Add `proptest` tests for `Best`


NicMcPhee:
At the moment the `Best` selector only has one static test, and we should add a `proptest` for it (which would be easy).

Tracked by #240

Issue #243, opened 2024-07-29
Add tests for `Weighted` selection


NicMcPhee:
This is going to be hard to test in a meaningful way since there's stochasticity at the level of which selection mechanism is chosen, _and_ in running that selection mechanism itself. Maybe it would make sense to implement a `Worst` selector as a mirror to `Best` (& there would be plausible uses for such a selector in EC work), and then we'd have two deterministic selectors, leaving all the randomness in the choice of the selector.

I don't know any non-statistical way to check that the weights are being handled correctly.

I had a look at [the testing for `choose_weighted` in the `rand` crate](https://docs.rs/rand/latest/src/rand/seq/mod.rs.html#1047), and it looks like they do 3,000 selections and "hope" that the total error across all those isn't "too high". They use a `test::rng` with a fixed seed, which probably gives them a little more control, but presumably they could make changes in how they generate their pseudo-random values that would alter their total error, so there's _some_ chance that their test will fail even if the logic is correct.

Tracked by #240

Issue #242, opened 2024-07-29
Add tests for `Random` selection


NicMcPhee:
I think our best bet is to use `proptest` to check that `Random` returns something from the population. As long as the population size is much smaller than the type space (like 20 values drawn from `i64`), then this would be at least somewhat informative.

This would pass if `Random` always returned, e.g., the first value in the population, which isn't awesome. I'm not sure how we'd deterministically test that it returns _random_ values, though.

Tracked by #240

Issue #240, opened 2024-07-28
Tracking issue for adding tests for selection mechanisms


NicMcPhee:
So far `Lexicase`, `Best`, and `Tournament` are the only selection mechanisms in `ec-core` that have any tests, and we should add/improve the tests for the others:

- [ ] #242 
- [ ] #243
- [ ] #244

We might want to build on the `proptest` strategies used in `Lexicase` to create some more general tooling to help with the tests. In particular a strategy that creates a random population with a known winning individual or set of winning individuals would be useful.

This should be seen as a tracking issue, and we should create separate issues/PRs for the specific changes.

PR #258, opened 2024-08-12
Add proptest tests for selectors

NicMcPhee:
This adds `proptest` for all the selectors mentioned in #240:

- #242
- #243 
- #244 

I also added a new `Worst` selector because that made it easier to write a reasonable test for the `Weighted` selector. That might also be useful in some EC contexts where, for example, we might want to select the worst individual for removal from the population.

We had talked about doing these in separate PRs, but in the end it was so straightforward that I figured I'd just do them all as a batch.

Closes: #240 
Closes: #242 
Closes: #243 
Closes: #244

### DIFF
--- a/packages/ec-core/src/operator/selector/best.rs
+++ b/packages/ec-core/src/operator/selector/best.rs
@@ -30,6 +30,8 @@ where
     )
 )]
 mod tests {
+    use test_strategy::proptest;
+
     use super::*;
 
     #[test]
@@ -38,5 +40,13 @@ mod tests {
         let mut rng = rand::thread_rng();
         assert_eq!(&9, Best.select(&pop, &mut rng).unwrap());
         assert_eq!(&9, Best.select(&pop, &mut rng).unwrap());
+    }
+
+    #[proptest]
+    fn test_best_select(#[any] values: [i32; 10]) {
+        let pop: Vec<i32> = values.into();
+        let mut rng = rand::thread_rng();
+        let largest = pop.iter().max().unwrap();
+        assert_eq!(largest, Best.select(&pop, &mut rng).unwrap());
     }
 }
--- a/packages/ec-core/src/operator/selector/best.rs
+++ b/packages/ec-core/src/operator/selector/best.rs
@@ -32,7 +32,7 @@ where
 mod tests {
     use test_strategy::proptest;
 
-    use super::*;
+    use super::{Best, Selector};
 
     #[test]
     fn can_select_twice() {
--- a/packages/ec-core/src/operator/selector/best.rs
+++ b/packages/ec-core/src/operator/selector/best.rs
@@ -36,6 +36,9 @@ mod tests {
 
     #[test]
     fn can_select_twice() {
+        // Currently `.select()` can't take an array, so we need to make this a `Vec`.
+        // Once we've generalized `.select()` appropriately we can change this to be
+        // an array. See #259
         let pop = vec![5, 8, 9, 6, 3, 2, 0];
         let mut rng = rand::thread_rng();
         assert_eq!(&9, Best.select(&pop, &mut rng).unwrap());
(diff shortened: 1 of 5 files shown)


packages/ec-core/src/operator/selector/best.rs
```diff
@@ -39,4 +41,12 @@ mod tests {
         assert_eq!(&9, Best.select(&pop, &mut rng).unwrap());
         assert_eq!(&9, Best.select(&pop, &mut rng).unwrap());
     }
+
+    #[proptest]
+    fn test_best_select(#[any] values: [i32; 10]) {
+        let pop: Vec<i32> = values.into();
```
JustusFluegel, line 47:
```suggestion
    fn test_best_select(#[map(|v: [i32;10]| v.into())] pop: Vec<i32>) {
```

packages/ec-core/src/operator/selector/random.rs
```diff
@@ -21,3 +21,26 @@ where
...
+    )
+)]
+mod tests {
+    use test_strategy::proptest;
+
+    use super::{Random, Selector};
+
+    #[proptest]
+    fn test_random(#[any] values: [i32; 10]) {
+        let pop: Vec<i32> = values.into();
```
JustusFluegel, line 41: same here

packages/ec-core/src/operator/selector/weighted.rs
```diff
@@ -48,3 +48,32 @@ where
...
+)]
+mod tests {
+    use test_strategy::proptest;
+
+    use super::Weighted;
+    use crate::operator::selector::{best::Best, worst::Worst, Selector};
+
+    #[proptest]
+    fn test_random(#[any] values: [i32; 10]) {
+        let pop: Vec<i32> = values.into();
```
JustusFluegel, line 69: and here

packages/ec-core/src/operator/selector/weighted.rs
```diff
@@ -48,3 +48,32 @@ where
...
+        let mut rng = rand::thread_rng();
+        // We'll make a selector that has a 50/50 chance of choosing the highest
+        // or lowest value.
+        let weighted = Weighted::new(Best, 1).with_selector(Worst, 1);
+        let selection = weighted.select(&pop, &mut rng).unwrap();
+        let smallest = values.iter().min().unwrap();
+        let largest = values.iter().max().unwrap();
```
JustusFluegel, line 76:
```suggestion
        let (smallest,largest) = values.iter().minmax().into_option().unwrap();
```
may be nicer, minmax is from itertools.

packages/ec-core/src/operator/selector/worst.rs
```diff
@@ -0,0 +1,52 @@
...
+    )
+)]
+mod tests {
+    use test_strategy::proptest;
+
+    use super::{Selector, Worst};
+
+    #[test]
+    fn can_select_twice() {
+        let pop = vec![5, 8, 9, 6, 3, 2, 10];
```
JustusFluegel, line 39: just use an array
```suggestion
        let pop = [5, 8, 9, 6, 3, 2, 10];
```

packages/ec-core/src/operator/selector/worst.rs
```diff
@@ -0,0 +1,52 @@
...
+        let mut rng = rand::thread_rng();
+        assert_eq!(&2, Worst.select(&pop, &mut rng).unwrap());
+        assert_eq!(&2, Worst.select(&pop, &mut rng).unwrap());
+    }
+
+    #[proptest]
+    fn test_worst_select(#[any] values: [i32; 10]) {
+        let pop: Vec<i32> = values.into();
```
JustusFluegel, line 47: same as above

JustusFluegel (dismissed):
Mostly looks good to me, I would personally clean up the vector generation like I added in the comment, take a look if you want to do that

packages/ec-core/src/operator/selector/worst.rs
NicMcPhee: This needs to remain a `Vec` because (currently) `.select()` doesn't work on arrays. Once that is fixed, we should change this.

packages/ec-core/src/operator/selector/worst.rs
NicMcPhee: Currently `.select()` can't take an array, so we need to make this a `Vec`. Once we've generalized `.select()` appropriately we can change this to be an array. See #259

JustusFluegel (approved):

PR #261, opened 2024-08-31
Add explicit errors for selectors

NicMcPhee:
Replace of "generic" `anyhow` errors with error types specific to the different selectors.

There is a general `SelectionError::EmptyPopulation` for the common error when trying to select from an empty population.

Different selection mechanisms have specific errors as necessary. `Tournament` has `TournamentError::LargerThanPopulation`, for example, for the case where the tournament size is larger than the population size.

### DIFF
--- a/packages/ec-core/src/operator/selector/tournament.rs
+++ b/packages/ec-core/src/operator/selector/tournament.rs
@@ -1,6 +1,6 @@
 use std::num::NonZeroUsize;
 
-use anyhow::{ensure, Context, Result};
+use miette::Diagnostic;
 use rand::{prelude::IndexedRandom, rngs::ThreadRng};
 
 use super::Selector;
@@ -8,6 +8,13 @@ use crate::population::Population;
 
 pub struct Tournament {
     size: NonZeroUsize,
+}
+
+#[derive(Debug, thiserror::Error, Diagnostic)]
+pub enum TournamentError {
+    #[error("Tournament size {0} was larger than population size {1}")]
+    #[diagnostic(help = "Ensure that the population has at least {0} individuals")]
+    LargerThanPopulation(usize, NonZeroUsize),
 }
 
 impl Tournament {
@@ -58,22 +65,27 @@ where
     P: Population + AsRef<[P::Individual]>,
     P::Individual: Ord,
 {
+    type Error = TournamentError;
+
     fn select<'pop>(
         &self,
         population: &'pop P,
         rng: &mut ThreadRng,
-    ) -> Result<&'pop P::Individual> {
-        ensure!(
-            population.size() >= self.size.into(),
-            "The population had size {} and we wanted a tournament of size {}",
-            population.size(),
-            self.size
-        );
+    ) -> Result<&'pop P::Individual, Self::Error> {
+        if population.size() < self.size.into() {
+            return Err(TournamentError::LargerThanPopulation(
+                population.size(),
+                self.size,
+            ));
+        }
         population
             .as_ref()
             .choose_multiple(rng, self.size.into())
             .max()
-            .with_context(|| format!("The tournament was empty; should have been {}", self.size))
+            // This should never happen, because an empty population will cause the
+            // `if` test test to return an `Err` since `self.size` is guaranteed to
+            // be greater than zero.
+            .ok_or_else(|| unreachable!("Population can't be empty here"))
     }
 }
 
@@ -94,7 +106,35 @@ mod tests {
     use test_strategy::proptest;
 
     use super::Tournament;
-    use crate::{individual::ec::EcIndividual, operator::selector::Selector};
+    use crate::{
+        individual::ec::EcIndividual,
+        operator::selector::{tournament::TournamentError, Selector},
+    };
+
+    #[test]
+    fn empty_population() {
+        let pop: Vec<i32> = Vec::new();
+        let mut rng = rand::thread_rng();
+        let selector = Tournament::new(NonZeroUsize::MIN);
+        assert!(matches!(
+            selector.select(&pop, &mut rng),
+            Err(TournamentError::LargerThanPopulation(0, NonZeroUsize::MIN))
+        ));
+    }
+
+    #[test]
+    fn tournament_size_larger_than_population() {
+        let pop: Vec<i32> = vec![0];
+        let mut rng = rand::thread_rng();
+        let selector = Tournament::of_size::<2>();
+        assert!(matches!(
+            selector.select(&pop, &mut rng),
+            Err(TournamentError::LargerThanPopulation(
+                1,
+               tournament_size
+            ))
+            if tournament_size == NonZeroUsize::new(2).unwrap()));
+    }
 
     #[test]
     fn tournament_size_1() {
--- a/packages/ec-core/Cargo.toml
+++ b/packages/ec-core/Cargo.toml
@@ -27,12 +27,13 @@ rand = { workspace = true, features = ["alloc"] }
 rayon = "1.7.0"
 macro_railroad_annotation = { workspace = true }
 rustversion = { workspace = true }
+thiserror = { workspace = true }
+miette = { workspace = true }
 
 [dev-dependencies]
 rand = { workspace = true, features = ["alloc", "small_rng"] }
 criterion = { workspace = true }
 miette = { workspace = true, features = ["fancy"] }
-thiserror = { workspace = true }
 proptest = { workspace = true }
 test-strategy = { workspace = true }
 
--- a/packages/ec-core/src/operator/selector/best.rs
+++ b/packages/ec-core/src/operator/selector/best.rs
@@ -1,7 +1,6 @@
-use anyhow::{Context, Result};
 use rand::rngs::ThreadRng;
 
-use super::Selector;
+use super::{error::SelectionError, Selector};
 use crate::population::Population;
 
 pub struct Best;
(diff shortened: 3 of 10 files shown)


JustusFluegel:
label police 🚨

packages/ec-core/src/operator/selector/weighted/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+pub mod weighted_selector_pair;
+pub mod with_weighted_selector;
+
+#[derive(Debug, thiserror::Error)]
+#[error("Overflow while trying to calculate the sum of the weights {0} and {1}.")]
+pub struct WeightSumOverflow(u32, u32);
+
+pub trait WithWeight {
+    fn weight(&self) -> u32;
+}
```
JustusFluegel, line 11: even if this isn't much I'd probably put this into it's own module

packages/ec-core/src/operator/selector/weighted/mod.rs
```diff
@@ -0,0 +1,57 @@
+pub mod weighted;
+pub mod weighted_selector_pair;
+pub mod with_weighted_selector;
+
+#[derive(Debug, thiserror::Error)]
+#[error("Overflow while trying to calculate the sum of the weights {0} and {1}.")]
+pub struct WeightSumOverflow(u32, u32);
```
JustusFluegel, line 7: I would either put this along the constructors (where the error occurs) or inside a seperate pub mod errors; - I would depend this on where we end up putting the other error.

packages/ec-core/src/operator/selector/weighted/mod.rs
```diff
@@ -0,0 +1,57 @@
+pub mod weighted;
+pub mod weighted_selector_pair;
+pub mod with_weighted_selector;
+
+#[derive(Debug, thiserror::Error)]
```
JustusFluegel, line 5: Also, this doesn't derive Diagnostic yet.

packages/ec-core/src/operator/selector/weighted/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+        // or lowest value.
+        let weighted = Weighted::new(Best, 1)
+            .with_selector_and_weight(Worst, 1)
+            .with_selector_and_weight(Random, 0)
+            .with_selector_and_weight(Tournament::of_size::<3>(), 2)
+            .unwrap();
+        let selection = weighted.select(&pop, &mut rng).unwrap();
+        assert!(pop.contains(selection));
+    }
+}
```
JustusFluegel, line 57: I would move the tests in the mod where we actually define the stuff we are testing, i.e. where the select impl  also is.

packages/ec-core/src/operator/selector/weighted/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+#[rustversion::attr(before(1.81), allow(clippy::unwrap_used))]
+#[rustversion::attr(
+    since(1.81),
+    expect(
+        clippy::unwrap_used,
+        reason = "Panicking is the best way to deal with errors in unit tests"
+    )
+)]
+mod tests {
+    use itertools::Itertools;
```
JustusFluegel, line 23: Also, we probably want tests for the error cases.

packages/ec-core/src/operator/selector/weighted/weighted_selector_pair.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Debug)]
+pub enum WeightedSelectorsError<A, B> {
+    A(A),
+    B(B),
+
+    ZeroWeightSum,
+
+    // #[error("Adding the weights {0} and {1} overflows")]
+    WeightSumOverflows(u32, u32),
+}
```
JustusFluegel, line 22: Depending on where we put this I'd create a errors module or leave it here - also, we need to implement Display, Error, and Diagnostic for this.

packages/ec-core/src/operator/selector/weighted/weighted_selector_pair.rs
```diff
@@ -0,0 +1,95 @@
...
+    ///   would overflow `u32::MAX`
+    pub fn new_with_weights(
+        a: A,
+        weight_a: u32,
+        b: B,
+        weight_b: u32,
+    ) -> Result<Self, WeightSumOverflow> {
+        Self::new(Weighted::new(a, weight_a), Weighted::new(b, weight_b))
+    }
+}
```
JustusFluegel, line 70: I don't know if we still really want this, after all we can now just write `Weighted::new(a, weight_a).with_selector_and_weight(b, weight_b)?` instead, and imho the args to this function are quite confusing.

packages/ec-core/src/operator/selector/weighted/weighted_selector_pair.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub fn new(a: A, b: B) -> Result<Self, WeightSumOverflow>
+    where
+        A: WithWeight,
+        B: WithWeight,
+    {
+        let a_weight = a.weight();
+        let b_weight = b.weight();
+        let Some(weight_sum) = a_weight.checked_add(b_weight) else {
+            return Err(WeightSumOverflow(a_weight, b_weight));
+        };
```
JustusFluegel, line 85:
```suggestion
        let weight_sum = a_weight.checked_add(b_weight).ok_or(WeightSumOverflow(a_weight, b_weight))?
```

JustusFluegel: Yes I know I wrote this myself. I also make mistakes :joy:

packages/ec-linear/examples/count_ones/main.rs
```diff
@@ -64,7 +64,7 @@ fn main() -> Result<()> {
 
     let num_test_cases = bit_length;
 
-    let selector = Weighted::new(Best, 1)
+    let selector = DynWeighted::new(Best, 1)
```
JustusFluegel, line 67: Question: Do we want `DynWeighted` or just `Weighted` in the examples?

packages/ec-linear/examples/hiff/main.rs
```diff
@@ -77,7 +77,7 @@ fn main() -> Result<()> {
 
     let num_test_cases = 2 * bit_length - 1;
 
-    let selector = Weighted::new(Best, 1)
+    let selector = DynWeighted::new(Best, 1)
```
JustusFluegel, line 80: Question: Do we want `DynWeighted` or just `Weighted` in the examples?

packages/push/examples/simple_regression/main.rs
```diff
@@ -143,7 +143,7 @@ fn main() -> Result<()> {
 
     let num_test_cases = 10;
 
-    let selector = Weighted::new(Best, 1)
+    let selector = DynWeighted::new(Best, 1)
```
JustusFluegel, line 146: Question: Do we want `DynWeighted` or just `Weighted` in the examples?

JustusFluegel:
Also, we should consider our module structure and whether we want to make some mods private and just re-export in the top-level weighted.rs

packages/ec-linear/examples/count_ones/main.rs
NicMcPhee: We'll address this in #263.

packages/push/examples/simple_regression/main.rs
NicMcPhee: We'll address this in #263.

packages/ec-linear/examples/hiff/main.rs
NicMcPhee: We'll address this in #263.

packages/ec-core/src/operator/selector/tournament.rs
```diff
@@ -1,15 +1,23 @@
...
     size: NonZeroUsize,
 }
 
+#[derive(Debug, thiserror::Error, Diagnostic)]
+pub enum TournamentError {
+    #[error("Tournament size {0} was larger than population size {1}")]
+    #[diagnostic(help = "Ensure that the population has at least {0} individuals")]
+    LargerThanPopulation(usize, NonZeroUsize),
+}
+
```
JustusFluegel, line 20: Since we don't have another error other than `LargerThanPopulation` and I don't see us changing the error paths of a TournamentSelector anytime soon, let's pull that out into a struct.

packages/ec-core/src/weighted/weighted_pair.rs
```diff
@@ -0,0 +1,179 @@
...
+    with_weight::WithWeight,
+};
+use crate::{operator::selector::Selector, population::Population};
+
+#[derive(Debug)]
+pub struct WeightedPair<A, B> {
+    pub(crate) a: A,
+    pub(crate) b: B,
+    pub(crate) distr: Option<Bernoulli>,
+    pub(crate) weight_sum: u32,
```
JustusFluegel, line 14: Just one thing I noticed, not that we necessarily need to change this, but we should consider it: Do we want the struct items ordered this way?

packages/ec-core/src/weighted/weighted_pair.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    #[test]
+    fn weight_sum_overflow_error_chaining() {
+        let weighted = Weighted::new(Best, u32::MAX).with_item_and_weight(Random, u32::MAX - 1);
+        assert!(weighted.is_err());
+        let weighted = weighted.with_item_and_weight(Worst, 0).unwrap_err();
+        assert_eq!(weighted, WeightSumOverflow(u32::MAX, u32::MAX - 1));
+    }
+}
```
JustusFluegel, line 179: Do we have a simple way to make the underlying selector error and add tests for that? If not we should maybe add a simple 

```rs
struct SelectorError;

struct ErroringSelector;

impl Selector for ErroringSelector {
    type Error = SelectorError;

    fn select(...) -> ... { SelectorError }
}
```
and add tests for that

JustusFluegel (dismissed):
LGTM, added a few comments but we can get this merged in if you think we shouldn't change anything there.

packages/ec-core/src/weighted/weighted_pair.rs
NicMcPhee: I'd be inclined to keep it this way.

packages/ec-core/src/weighted/weighted_pair.rs
NicMcPhee: I moved this to a new issue: #269

JustusFluegel (approved):

PR #262, opened 2024-09-01
Less restrictive weighted generics

JustusFluegel:
As discussed on stream here a version of the recursive weighted selector where I've tried to move away from generics on the structs as much as possible.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -5,5 +5,15 @@ allowed-duplicate-crates = [
   "base64",
   "itertools",
   "syn",
-  "heck"
+  "heck",
+  # https://github.com/eminence/terminal-size/pull/62
+  "windows-sys",
+  "windows-targets",
+  "windows_aarch64_gnullvm",
+  "windows_aarch64_msvc",
+  "windows_i686_gnu",
+  "windows_i686_msvc",
+  "windows_x86_64_gnu",
+  "windows_x86_64_gnullvm",
+  "windows_x86_64_msvc"
 ]
--- a/packages/ec-core/src/operator/selector/recursive_weighted.rs
+++ b/packages/ec-core/src/operator/selector/recursive_weighted.rs
@@ -1,72 +1,115 @@
-use std::marker::PhantomData;
+use rand::distr::{Bernoulli, Distribution};
 
-use rand::Rng;
-
-use super::{weighted, Selector};
+use super::Selector;
 use crate::population::Population;
 
-#[derive(Debug)]
-pub enum WeightedSelectorCreationError {
-    WeightSumOverflows(usize, usize),
-}
+#[derive(Debug, thiserror::Error)]
+#[error("Overflow while trying to calculate the sum of the weights {0} and {1}.")]
+pub struct WeightSumOverflow(u32, u32);
 
-pub trait WeightedSelector<P>: Selector<P>
-where
-    P: Population,
-{
-    fn weight(&self) -> usize;
+pub trait WeightedSelector {
+    fn weight(&self) -> u32;
 }
 
 pub struct Weighted<T> {
     selector: T,
-    weight: usize,
+    weight: u32,
+}
+
+pub trait WithWeightedSelector
+where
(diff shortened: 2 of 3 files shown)


packages/ec-core/src/operator/selector/recursive_weighted.rs
```diff
@@ -1,72 +1,115 @@
...
 }
 
-pub trait WeightedSelector<P>: Selector<P>
+pub trait WithWeightedSelector
 where
-    P: Population,
+    Self: Sized,
 {
-    fn weight(&self) -> usize;
+    type Inner;
```
NicMcPhee, line 23: I don't think that's a great name. It doesn't tell us anything about what kind of thing `Inner` probably is.

packages/ec-core/src/operator/selector/recursive_weighted.rs
```diff
@@ -87,154 +130,95 @@ where
...
-    A: WeightedSelector<P>,
-    B: WeightedSelector<P>,
-{
+pub struct WeightedSelectorPair<A, B> {
     a: A,
     b: B,
-    // weight: usize,
-    _p: PhantomData<P>,
+    distr: Option<Bernoulli>,
+    weight_sum: u32,
```
NicMcPhee, line 137: Remind me again why you prefer `u32` over `usize` here? I'm not sure I care, but I am curious.

packages/ec-core/src/operator/selector/recursive_weighted.rs
```diff
@@ -87,154 +130,95 @@ where
...
     ) -> Result<&'pop <P as Population>::Individual, Self::Error> {
-        let a_weight = self.a.weight();
-        let b_weight = self.b.weight();
-        // For performance reasons, it would be nice to check that the sum works
-        // and is > 0 at construction time so we don't have to do it here.
-        let weight_sum = a_weight
-            .checked_add(b_weight)
-            .ok_or_else(|| WeightedSelectorsError::WeightSumOverflows(a_weight, b_weight))?;
-        if weight_sum == 0 {
+        let Some(distr) = self.distr else {
```
NicMcPhee, line 164: Doesn't this want to return something different in the `else` case?

NicMcPhee (commented):
I _think_ this looks reasonable, but I'm not sure I understand the motivation behind all the parts. Let's talk about it in today's stream.

clippy.toml
```diff
@@ -5,5 +5,15 @@ allowed-duplicate-crates = [
   "base64",
   "itertools",
   "syn",
-  "heck"
+  "heck",
+  # https://github.com/eminence/terminal-size/pull/62
+  "windows-sys",
```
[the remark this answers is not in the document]
JustusFluegel: windows-sys is a indirect dependency through clap and miette which use that to change how they format their output depending on terminal size, on windows. That's why it's in the tree and that's why the lint reminded us that we have it duplicated (cuz terminal-size hasn't updated it yet)

packages/ec-core/src/operator/selector/recursive_weighted.rs
JustusFluegel: maybe rename it to InnerSelector?

packages/ec-core/src/operator/selector/recursive_weighted.rs
JustusFluegel: using `u32` over `usize` is probably more than we need ( we don't really need `u64` on 64-bit platforms and weights aren't really dependent on memory size or pointer width or anything so using usize in the first place doesn't really make sense - also using u32 allows us to use the rand bernouli distribution

packages/ec-core/src/operator/selector/recursive_weighted.rs
JustusFluegel: The distr is only not set in the constructor in the zero sum case tho so that error makes sense here imho.

NicMcPhee (approved):

PR #276, opened 2024-10-22
Elide lifetimes

JustusFluegel:
Fixes new clippy lints on nightly & beta.

### DIFF
--- a/packages/push/src/push_vm/stack.rs
+++ b/packages/push/src/push_vm/stack.rs
@@ -411,10 +411,10 @@ impl<T> Stack<T> {
     {
         let iter = iter.into_iter();
         // Check that adding these items won't overflow the stack.
-        if !iter
+        if iter
             .len()
             .checked_add(self.size())
-            .is_some_and(|x| x <= self.max_stack_size)
+            .is_none_or(|x| x > self.max_stack_size)
         {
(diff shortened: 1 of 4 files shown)


packages/push/src/push_vm/stack.rs
```diff
@@ -411,10 +411,10 @@ impl<T> Stack<T> {
     {
         let iter = iter.into_iter();
         // Check that adding these items won't overflow the stack.
-        if !iter
+        if iter
             .len()
             .checked_add(self.size())
-            .is_some_and(|x| x <= self.max_stack_size)
+            .is_none_or(|x| x > self.max_stack_size)
```
NicMcPhee, line 417: Nice – that's definitely much better!

NicMcPhee (approved):
Nice – thanks!

packages/push/src/push_vm/stack.rs
JustusFluegel: Yeah that is what that does, they kind of depend on each other, #277 is based on top of this branch so we can either merge them chronologically or we can just merge #277  and then close this as well.

NicMcPhee:
These changes were part of #277, right? So we can/should just close this one?

PR #277, opened 2024-10-22
Rustfmt changes 1.81

JustusFluegel:
This applies the changes done in rustfmt for rust 1.81, which is currently incompatible with stable (till the 17th of November), and should be merged after that.

Also, this is based on #276


NicMcPhee (approved):
This looks great. Thanks for putting all this together for us.

PR #290, opened 2024-11-24
Rename `Identity` operator to `Constant`; implement `Identity`

### LINKED ISSUE #288: Rename `Identity` to `Constant`; add actual `Identity` implementation
My implementation of the `Identity` operator is actually a constant operator, i.e., it ignores its input and always returns the same value. So its name is wrong and should be `Constant`.

There might be some value in having an `Identity` operator, however, so we should perhaps implement that as well.

NicMcPhee:
The old implementation of `Identity` was really a constant operator (it always returned the same value regardless of its input), so this renames that and introduces an implementation of `Identity` that actually returns its input.

Still need to:

- [x] Document these operators
- [x] Test these operators

Closes #288


JustusFluegel (approved):
Don't think I have anything to complain about here ( looks good ) apart from maybe making the constant test a `#[test_case(...)]` test and testing more than just a single constant integer? More of a nit tho, so I'll approve this 👍

NicMcPhee:
Actually, I think the real improvement to be made there would be to switch to (or add) a `proptest` because then the integers would be randomly generated each time, which would provide richer coverage than anything we could write by hand.

That said, this is pretty straightforward code, so I'm happy just moving forward.

Issue #64, opened 2024-02-04
Move away from `anyhow::Error` inside lib code


JustusFluegel:
Adding explicit errors is helpful for library consumers, `anyhow::Error` doesn't really provide any indication on a programmatic level of WHY it failed,  so it becomes impossible to "catch" a specific error and recover from that. 

Maybe consider using miette for prettier diagnostics as well?

---

## Tracking

We should generate a list of sections of the code where this change needs to happen so we can mark them as completed as we work through them.

### `ec-core`

- `operator`
   - `selector`
      - [x] `Best`
      - [x] `Lexicase`
      - [x] `Random`
      - [x] `Tournament`
      - [x] `Weighted`
   - `mutator`
   - `recombinator`
   - `composable`
      - [ ] `and`
      - [ ] `map`
      - [ ] `repeat_with`
      - [ ] `then`

NicMcPhee:
I've never use miette, but it's on my list of things it would be good to learn, so I'm fine with exploring that.

NicMcPhee:
There's a _lot_ to do here, and I'd rather not have this end up in an enormous PR that's impossible to review. So why don't we treat this as a tracking issue, and do the work in little PRs that we record here.

PR #305, opened 2024-12-16
Migrate away from anyhow Errors and to custom error types

JustusFluegel:
This migrates the remaining code inside the lib crates away from `anyhow`.
As part of the migration I've also ran `cargo update` to update the dependencies to the latest semver-compatible versions, so the following changes might also be observed inside this pr:
- Migrating away from deprecated `rand` code
- Updating the Cargo.lock

Notes / Fixmes (Create issues for this if not fixed by time of merging this pr), checkmarks indicate fixed:

- [x] Potentially breaking: This currently "dealt with" error types in the `map` operator by just returning the error from the applied operator directly. This isn't ideal, as this loses context we had before (using .context) of where the error occurred. Instead, create a simple newtype error wrapper (with a  '#[from]` impl) for the error which adds some context.

### DIFF
--- a/packages/ec-core/src/operator/composable/map.rs
+++ b/packages/ec-core/src/operator/composable/map.rs
@@ -56,7 +56,7 @@ where
         rng: &mut rand::rngs::ThreadRng,
     ) -> Result<Self::Output, Self::Error> {
         let first_result = self.f.apply(x, rng).map_err(|e| MapError(e, 0))?;
-        let second_result = self.f.apply(y, rng).map_err(|e| MapError(e, 0))?;
+        let second_result = self.f.apply(y, rng).map_err(|e| MapError(e, 1))?;
         Ok([first_result, second_result])
     }
 }
@@ -74,7 +74,7 @@ where
         rng: &mut rand::rngs::ThreadRng,
     ) -> Result<Self::Output, Self::Error> {
         let first_result = self.f.apply(x, rng).map_err(|e| MapError(e, 0))?;
-        let second_result = self.f.apply(y, rng).map_err(|e| MapError(e, 0))?;
+        let second_result = self.f.apply(y, rng).map_err(|e| MapError(e, 1))?;
         Ok((first_result, second_result))
     }
 }
--- a/packages/ec-core/src/operator/composable/map.rs
+++ b/packages/ec-core/src/operator/composable/map.rs
@@ -1,5 +1,3 @@
-use anyhow::Context;
-
 use super::Composable;
 use crate::operator::Operator;
 
@@ -13,30 +11,52 @@ impl<F> Map<F> {
     }
 }
 
+#[derive(Debug)]
+pub struct MapError<T>(T, usize);
+
+impl<T> std::fmt::Display for MapError<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "Error while applying passed operator on the {}-th element of the mapped iterable",
+            self.1
+        )
+    }
+}
+
+impl<T> std::error::Error for MapError<T>
+where
+    T: std::error::Error + 'static,
+{
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(&self.0)
+    }
+}
+
+impl<T> miette::Diagnostic for MapError<T>
+where
+    T: miette::Diagnostic + 'static,
+{
+    fn diagnostic_source(&self) -> Option<&dyn miette::Diagnostic> {
+        Some(&self.0)
+    }
+}
+
 // I think I can parameterize over the 2 here to make this more general?
 impl<F, Input> Operator<[Input; 2]> for Map<F>
 where
     F: Operator<Input>,
-    anyhow::Error: From<F::Error>,
 {
     type Output = [F::Output; 2];
-    type Error = anyhow::Error;
+    type Error = MapError<F::Error>;
 
     fn apply(
         &self,
         [x, y]: [Input; 2],
         rng: &mut rand::rngs::ThreadRng,
     ) -> Result<Self::Output, Self::Error> {
-        let first_result = self
-            .f
-            .apply(x, rng)
-            .map_err(anyhow::Error::from)
-            .context("Calling f with {x} in `Map` failed")?;
-        let second_result = self
-            .f
-            .apply(y, rng)
-            .map_err(anyhow::Error::from)
-            .context("Calling f with {y} in `Map` failed")?;
+        let first_result = self.f.apply(x, rng).map_err(|e| MapError(e, 0))?;
+        let second_result = self.f.apply(y, rng).map_err(|e| MapError(e, 0))?;
         Ok([first_result, second_result])
     }
 }
@@ -44,26 +64,17 @@ where
 impl<F, Input> Operator<(Input, Input)> for Map<F>
 where
     F: Operator<Input>,
-    anyhow::Error: From<F::Error>,
 {
     type Output = (F::Output, F::Output);
-    type Error = anyhow::Error;
+    type Error = MapError<F::Error>;
 
     fn apply(
         &self,
         (x, y): (Input, Input),
         rng: &mut rand::rngs::ThreadRng,
     ) -> Result<Self::Output, Self::Error> {
-        let first_result = self
-            .f
-            .apply(x, rng)
-            .map_err(anyhow::Error::from)
-            .context("Calling f with {x} in `Map` failed")?;
-        let second_result = self
-            .f
-            .apply(y, rng)
-            .map_err(anyhow::Error::from)
-            .context("Calling f with {y} in `Map` failed")?;
+        let first_result = self.f.apply(x, rng).map_err(|e| MapError(e, 0))?;
+        let second_result = self.f.apply(y, rng).map_err(|e| MapError(e, 0))?;
         Ok((first_result, second_result))
     }
 }
@@ -71,10 +82,9 @@ where
 impl<F, Input> Operator<Vec<Input>> for Map<F>
 where
     F: Operator<Input>,
-    anyhow::Error: From<F::Error>,
 {
     type Output = Vec<F::Output>;
-    type Error = anyhow::Error;
+    type Error = MapError<F::Error>;
 
     fn apply(
         &self,
@@ -83,12 +93,8 @@ where
     ) -> Result<Self::Output, Self::Error> {
         input
             .into_iter()
-            .map(|x| {
-                self.f
-                    .apply(x, rng)
-                    .map_err(anyhow::Error::from)
-                    .context("Applying f to {x} in `Map` failed")
-            })
+            .enumerate()
+            .map(|(i, x)| self.f.apply(x, rng).map_err(|e| MapError(e, i)))
             .collect()
     }
 }
--- a/packages/ec-core/src/operator/composable/repeat_with.rs
+++ b/packages/ec-core/src/operator/composable/repeat_with.rs
@@ -59,7 +59,7 @@ impl<F, const N: usize> Composable for RepeatWith<F, N> {}
 mod tests {
     use std::{convert::Infallible, ops::Range};
 
-    use rand::{Rng, thread_rng};
+    use rand::{Rng, rng};
 
     use super::*;
 
@@ -82,7 +82,7 @@ mod tests {
     fn deterministic() {
         const LENGTH: usize = 5;
         let desired_value = 7;
-        let mut rng = thread_rng();
+        let mut rng = rng();
         let repeater: RepeatWith<AddOne, LENGTH> = RepeatWith::new(AddOne);
         let result = repeater.apply(desired_value, &mut rng).unwrap();
         assert_eq!(LENGTH, result.len());
@@ -99,7 +99,7 @@ mod tests {
             range: Range<i32>,
             rng: &mut rand::rngs::ThreadRng,
         ) -> Result<Self::Output, Self::Error> {
-            Ok(rng.gen_range(range))
+            Ok(rng.random_range(range))
         }
     }
     impl Composable for UniformRange {}
@@ -108,7 +108,7 @@ mod tests {
     fn stochastic() {
         const LENGTH: usize = 5;
         let range = 0..7;
-        let mut rng = thread_rng();
+        let mut rng = rng();
         let repeater: RepeatWith<UniformRange, LENGTH> = RepeatWith::new(UniformRange);
         let result = repeater.apply(range.clone(), &mut rng).unwrap();
         assert_eq!(LENGTH, result.len());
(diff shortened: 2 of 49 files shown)


JustusFluegel:
This actually ended up being a very straightforward pr to implement, I just searched for anyhow mentions and either removed the path or the import and just ended up fixing the diagnostics, for the most part. The diff stats are also pretty reasonable if you exclude the lock file:

```
 Cargo.toml                                                   |   2 +-
 clippy.toml                                                  |  13 ++++-------
 packages/ec-core/Cargo.toml                                  |   2 +-
 packages/ec-core/benches/test_results.rs                     |   8 +++----
 packages/ec-core/src/child_maker.rs                          |  39 +++++++++++--------------------
 packages/ec-core/src/distributions/collection.rs             |   5 ----
 packages/ec-core/src/distributions/wrappers/owned.rs         |   2 +-
 packages/ec-core/src/distributions/wrappers/slice_cloning.rs |  16 ++++---------
 packages/ec-core/src/generation.rs                           |  72 +++++++++++++++++++++++++++++++++++++++-----------------
 packages/ec-core/src/operator/composable/and.rs              |  61 ++++++++++++++++++++++++++++++++++++-----------
 packages/ec-core/src/operator/composable/map.rs              |  74 ++++++++++++++++++++++++++++++---------------------------
 packages/ec-core/src/operator/composable/repeat_with.rs      |  17 +++++---------
 packages/ec-core/src/operator/composable/then.rs             |  61 ++++++++++++++++++++++++++++++++++++++---------
 packages/ec-core/src/operator/constant.rs                    |   8 +++----
 packages/ec-core/src/operator/genome_extractor.rs            |  25 +++++++++++++-------
 packages/ec-core/src/operator/identity.rs                    |   8 +++----
 packages/ec-core/src/operator/mutator/mod.rs                 |  92 +++++++++++++++++++++++++++++++++++++++++++++++------------------------
 packages/ec-core/src/operator/recombinator/mod.rs            |  82 ++++++++++++++++++++++++++++++++++++----------------------------
 packages/ec-core/src/operator/selector/best.rs               |   6 ++---
 packages/ec-core/src/operator/selector/dyn_weighted.rs       |   2 +-
 packages/ec-core/src/operator/selector/lexicase.rs           |  12 +++++-----
 packages/ec-core/src/operator/selector/mod.rs                | 117 ++++++++++++++++++++++++++++++++++++++++++++++++++-----------------------------------------
 packages/ec-core/src/operator/selector/random.rs             |   4 ++--
 packages/ec-core/src/operator/selector/tournament.rs         |  12 +++++-----
 packages/ec-core/src/operator/selector/worst.rs              |   6 ++---
 packages/ec-core/src/population.rs                           |   6 ++---
 packages/ec-core/src/weighted/weighted_pair.rs               |   6 ++---
 packages/ec-linear/Cargo.toml                                |   5 +++-
 packages/ec-linear/examples/count_ones/main.rs               |  19 ++++++++-------
 packages/ec-linear/examples/hiff/main.rs                     |  17 ++++++++------
 packages/ec-linear/src/genome/bitstring.rs                   |  36 +++++++++++++++++++++-------
 packages/ec-linear/src/mutator/umad.rs                       |  18 ++++++++------
 packages/ec-linear/src/mutator/with_one_over_length.rs       |  47 ++++++++++++++++++++++---------------
 packages/ec-linear/src/mutator/with_rate.rs                  |  18 +++++++-------
 packages/ec-linear/src/recombinator/crossover.rs             |  23 ++++++++++++------
 packages/ec-linear/src/recombinator/errors.rs                |  95 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 packages/ec-linear/src/recombinator/mod.rs                   |   1 +
 packages/ec-linear/src/recombinator/two_point_xo.rs          |  53 +++++++++++++++++++++++------------------
 packages/ec-linear/src/recombinator/uniform_xo.rs            |  44 +++++++++++++++++++---------------
 packages/push/Cargo.toml                                     |   3 ++-
 packages/push/examples/complex_regression/main.rs            |  15 ++++++------
 packages/push/examples/median/main.rs                        |  18 +++++++-------
 packages/push/examples/simple_regression/main.rs             |  15 ++++++------
 packages/push/examples/smallest/main.rs                      |  18 +++++++-------
 packages/push/src/genome/plushy.rs                           |  10 +++-----
 packages/push/src/instruction/instruction_error.rs           |  17 +++++++++++---
 packages/push/src/instruction/int/mod.rs                     |   7 +++++-
 packages/push/src/push_vm/stack.rs                           |   5 +++-
 48 files changed, 779 insertions(+), 463 deletions(-)
```

If someone ( @NicMcPhee ) has the time to review this I would appreciate it greatly :)

JustusFluegel:
And fun fact: This actually showed us a place in the tests where the error was uninhabited and the `.unwrap()` could be removed and replaced with a `let Ok(value) = ...` (without an `else` clause) - This was more of an side effect of the expect lint expectation for the unwrap lint now being unfulfilled since it doesn't trigger the lint for unwraps on uninhabited errors, so we should probably manually go through the example/test code and try to fix that wherever possible.

JustusFluegel:
Another fun example I found of the error types being useful is the following:

In the `complex_regression` example if we take a look at the error type returned by `serial_next()` in the generation loop, we find the following:

`ThenError<ThenError<EmptyPopulation, Infallible>, Infallible>` which clearly shows us that the only possible error that can happen in this case results from an empty input population! Quite slick, if you ask me.

JustusFluegel:
Closes #64

NicMcPhee (commented):
Thanks for doing all this work!

I'm only about 1/3 of the way through the files, but so far most of it looks awesome. I did find what I think are two mistakes, and I have one question that's more stylistic in nature.

I think that in a perfect world it would have made more sense to separate the error handling and the `thread_rng` changes into two separate PRs, although I also appreciate your desire to just get both of those things done here.

I'll try to get through the rest tomorrow before the stream.

packages/ec-core/src/operator/composable/repeat_with.rs
```diff
@@ -82,9 +77,9 @@ mod tests {
     fn deterministic() {
         const LENGTH: usize = 5;
         let desired_value = 7;
-        let mut rng = thread_rng();
+        let mut rng = rng();
         let repeater: RepeatWith<AddOne, LENGTH> = RepeatWith::new(AddOne);
-        let result = repeater.apply(desired_value, &mut rng).unwrap();
+        let Ok(result) = repeater.apply(desired_value, &mut rng);
```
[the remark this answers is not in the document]
JustusFluegel: It can't panic, that's the thing - the compiler verified at compile time that the error variant Is uninhabited (i.e never constructed, a enum without variants, a type you can't actually construct) else the compiler wouldn't let you write this. Think of it as the compiler verifying that the type is `Result<T,, Infallible>`

JustusFluegel:
Well the only Rng changes here is updating the rand package, I think I agree now that I did that that we probably should have done that in a separate pr. But the changes here do not actually change any functionality, they just use the newly renamed functions

packages/ec-core/src/operator/composable/map.rs
```diff
@@ -13,68 +11,80 @@ impl<F> Map<F> {
...
-            .apply(x, rng)
-            .map_err(anyhow::Error::from)
-            .context("Calling f with {x} in `Map` failed")?;
-        let second_result = self
-            .f
-            .apply(y, rng)
-            .map_err(anyhow::Error::from)
-            .context("Calling f with {y} in `Map` failed")?;
+        let first_result = self.f.apply(x, rng).map_err(|e| MapError(e, 0))?;
+        let second_result = self.f.apply(y, rng).map_err(|e| MapError(e, 0))?;
```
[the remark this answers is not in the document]
JustusFluegel: Totally, thanks for catching that

packages/ec-core/src/operator/composable/map.rs
```diff
@@ -13,68 +11,80 @@ impl<F> Map<F> {
...
```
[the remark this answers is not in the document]
JustusFluegel: Totally, thanks for catching that

packages/ec-core/src/operator/composable/repeat_with.rs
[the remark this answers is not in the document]
JustusFluegel: Think of it like the compiler narrowing the type to

```rust
enum Result<T,E> {
    Ok(T)
}
```

packages/ec-core/src/operator/selector/mod.rs
```diff
@@ -289,26 +299,29 @@ where
...
 
     use super::{Select, Selector};
     use crate::operator::{Composable, Operator};
 
     // A simple `Selector`` that always returns the first item in a vector.
     struct First;
 
+    #[derive(Debug, thiserror::Error)]
+    #[error("At least one individual in the population is required to make a selection from it")]
+    struct EmptyPopulationError;
```
NicMcPhee, line 314: Is there a reason to create a new `EmptyPopulationError` in the various examples and tests in this file instead of using the `EmptyPopulation` error type that is used in the actual selectors?

Having `EmptyPopulationError` does illustrate how one might create an error type when making a new selector, but using the existing error type would highlight its existence and encourage people to use that instead of making a new one.

Thoughts?

packages/ec-linear/src/genome/bitstring.rs
```diff
@@ -129,28 +131,46 @@ impl Linear for Bitstring {
     }
 }
 
+#[derive(Debug, thiserror::Error, Diagnostic)]
+#[error("Failed to access gene at position {0}")]
+#[diagnostic(help = "Check that the gene is long enough")]
```
NicMcPhee, line 136: I would guess that it's more common that the index is out of range rather than the gene is "too short". What sort of error does `.get()` provide when you try to access an element out of range on something like a `Vec`?

packages/ec-linear/src/genome/bitstring.rs
```diff
@@ -129,28 +131,46 @@ impl Linear for Bitstring {
...
+pub struct GeneAccess(usize);
+
+#[derive(Debug, thiserror::Error, Diagnostic)]
+#[error("Failed to access gene at range {}..{}", self.0.start, self.0.end)]
+#[diagnostic(help = "Check that the gene is long enough")]
```
NicMcPhee, line 141: Same question, basically.

packages/ec-linear/src/mutator/with_one_over_length.rs
```diff
@@ -10,20 +10,27 @@ use crate::genome::Linear;
...
+        let genome_length = genome
+            .len()
+            .to_f32()
+            .ok_or(GenomeSizeConversionError(genome.len()))?;
         let mutation_rate = 1.0 / genome_length;
         let mutator = WithRate::new(mutation_rate);
-        mutator.mutate(genome, rng)
+        mutator
+            .mutate(genome, rng)
+            .map_err(|_: Infallible| unreachable!())
```
NicMcPhee, line 33: This seems rather subtle to me, and perhaps deserving of a comment? I know I had to stop and think about it for a second.

packages/ec-linear/src/mutator/with_one_over_length.rs
```diff
@@ -32,16 +39,18 @@ where
...
+        let genome_length = genome
+            .size()
+            .to_f32()
+            .ok_or_else(|| GenomeSizeConversionError(genome.size()))?;
         let mutation_rate = 1.0 / genome_length;
         let mutator = WithRate::new(mutation_rate);
-        mutator.mutate(genome, rng)
+        mutator
+            .mutate(genome, rng)
+            .map_err(|_: Infallible| unreachable!())
```
NicMcPhee, line 53: And again.

packages/ec-linear/src/recombinator/errors.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Debug, thiserror::Error, Diagnostic)]
+#[error("Attempted to perform TwoPointXo on genomes of different lengths {0} and {1}")]
+#[diagnostic(help = "Ensure your genomes are of uniform length")]
+pub struct DifferentGenomeLength(pub usize, pub usize);
+
+#[derive(Debug)]
+pub enum CrossoverGeneError<E> {
+    DifferentGenomeLength(DifferentGenomeLength),
+    Crossover(E),
+}
```
NicMcPhee, line 17: I feel like some sort of comment is warranted here explaining what the `Crossover` variant is doing. This is especially true since the rest of this file (nearly 100 lines) is really about dealing with the fact that we couldn't just use things like `transparent` here, but had implement things like `source` ourselves.

packages/push/src/instruction/instruction_error.rs
```diff
@@ -1,15 +1,26 @@
...
     /// Stack errors can be things like stack over- or underflows.
     #[error(transparent)]
-    StackError(#[from] StackError),
+    StackError(
+        #[from]
+        #[diagnostic_source]
+        StackError,
+    ),
     #[error("Exceeded the maximum step limit {step_limit}")]
+    #[diagnostic(help = "You might want to increase your step limit.")]
```
NicMcPhee, line 17: The real issue here is probably that there's an infinite loop, or that evolution has created some _really_ big data structure (like an enormous string) that it's then trying to process.

If you get a `StepLimitExceeded` then it's probably not a "hard" error that would ever make it to the user. Instead it's something that the Push interpreter would process, usually by terminating the execution of this program and (when possible) returning whatever value is on top of the appropriate stack.

packages/push/src/push_vm/stack.rs
```diff
@@ -83,14 +84,16 @@ pub trait HasStack<T> {
...
+#[derive(thiserror::Error, Debug, Eq, PartialEq, Diagnostic)]
 pub enum StackError {
     #[error("Requested {num_requested} elements from stack with {num_present} elements.")]
+    #[diagnostic(severity(Warning))]
     Underflow {
         num_requested: usize,
         num_present: usize,
     },
     #[error("Pushed onto full stack of type {stack_type}.")]
+    #[diagnostic(help = "You might want to increase your stack size", severity(Warning))]
```
NicMcPhee, line 96: Similar to the comment above, this is problem a symptom of a runaway program of some sort, and it's unlikely that (a) the user will ever see this and (b) that increasing the stack size will really help.

NicMcPhee (commented):
Thanks _so_ much for all this work!

All my comments are small, and we can discuss them in today's stream if you can make it.

packages/ec-core/src/operator/composable/repeat_with.rs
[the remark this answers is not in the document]
NicMcPhee: That's super cool, and wasn't not obvious to me. So a comment is still probably a good idea, but that's really nifty.

packages/ec-core/src/operator/selector/mod.rs
NicMcPhee: I'm inclined to use the existing error type instead of creating a new one.

NicMcPhee (approved):
This looks wonderful – thanks!

PR #307, opened 2024-12-25
Move from static `ThreadRng` parameter to generic `R: Rng + ?Sized`.

JustusFluegel:
This removes all the hardcoded `ThreadRng` parameters in our library code. 

~~Since those traits might now loose Object-Safety, we need erased variants for them. This doesn't add any Erased traits for now, see the next commit for that.~~ Edit: now included

This also adds a feature `erased` to all crates that need type erased traits, which can be used to optionally enable them. Since we kind of hand-rolled an erased type for the DynWeighted Selector previously, this moves the DynWeighted selector behind that feature flag and uses our newly created erased variant of selector as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -345,6 +345,7 @@ name = "ec-core"
 version = "0.1.0"
 dependencies = [
  "criterion",
+ "ec_macros",
  "itertools 0.13.0",
  "macro_railroad_annotation",
  "miette",
@@ -369,6 +370,16 @@ dependencies = [
  "rand 0.9.0-beta.1",
  "static_assertions",
  "thiserror",
+]
+
+[[package]]
+name = "ec_macros"
+version = "0.1.0"
+dependencies = [
+ "manyhow",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.90",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,6 +353,7 @@ dependencies = [
  "proptest",
  "rand 0.9.0-beta.1",
  "rayon",
+ "static_assertions",
  "test-strategy",
  "thiserror",
 ]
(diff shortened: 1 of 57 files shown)


JustusFluegel:
This pr will probably have a few merge conflicts with #305 so we should probably get that merged in first and then I can deal with those here, once this isn't in a draft state anymore

JustusFluegel:
So this should be mostly good to go except that I still want to think about the organization of the erased traits (they should maybe be in a `mod erased; pub use erased::*;` instead so that we don't need to repeat `#[cfg(feature=erased)]` everywhere) and how I did the normal-trait-impl on trait objects, I am not quite happy with creating a macro for each one yet and repeating that everywhere, especially also repeating a quasi-identical invocation of that macro everywhere. I probably want to create a decl macro to be able to write something along the lines of

```rs
erased_impl!{
     impl<T> SomeTrait<T> for dyn DynSomeTrait<T> {
        ...
     }
};
```

which will then also add the additional trait impls without having to create a new macro for each dyn trait.

JustusFluegel:
Ok I did what I talked about in the last comment but it turned out that a decl macro while definitely doable was reaaaly unreadable and not at all nice to write so I did a super simple proc macro instead which solved all of that.

JustusFluegel:
This pr should now be mostly ready to merge in but since it most definitely will have merge conflicts with #305 I'll leave this as a draft for now (even if it doesn't have merge conflicts it should fail to compile after the merge since I just copied anyhow stuff in this branch for now, which needs to also be replaced with the new assoc error types - should be pretty straightforward tho and not more than 20 minutes of work.

JustusFluegel:
Reorganized the commits a bit they were a bit messed up

NicMcPhee (commented):
It would be nice to have some documentation on the erased/dyn compatible versions of things like `ChildMaker` indicating what the trade-offs are and why you would prefer one to the other in particular circumstances.

NicMcPhee (approved):

PR #308, opened 2024-12-27
Implement Population trait via a blanket impl for any objects that have a sized iterator from it's reference and not just for `Vec<T>`

JustusFluegel:
Also removes some superfluous collects using iterator equality and some unneeded vecs in other places

### DIFF
--- a/packages/ec-core/src/population.rs
+++ b/packages/ec-core/src/population.rs
@@ -12,6 +12,17 @@ impl<T, I> Population for T
 where
     // An alternative would be `T: Deref<[T]>` but this also supports
     // hashsets, etc.
+    //
+    // This might look a bit confusing but essentially what this bound
+    // combination does is require both a `iter(&self) -> impl Iterator<Item
+    // = &T>` as well as a `into_iter(self) -> impl Iterator<Item = T>`
+    // method on the collection implementing population,
+    // which enables us to get the size of the collection without consuming
+    // it (through the `.iter()` method, since it takes `&self`) as well
+    // as turning this population into an iterator.
+    //
+    // Just requiring `.into_iter()` wouldn't work, since we don't want to
+    // consume the population to implement the `.size()` method.
     for<'a> &'a T: IntoIterator<Item = &'a I, IntoIter: ExactSizeIterator>,
     T: IntoIterator<Item = I>,
 {
--- a/packages/ec-core/src/population.rs
+++ b/packages/ec-core/src/population.rs
@@ -8,11 +8,17 @@ pub trait Population {
     fn size(&self) -> usize;
 }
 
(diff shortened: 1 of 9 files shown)


JustusFluegel:
Nightly ci fail will be fixed by #305

JustusFluegel:
(accidentally force pushed to zero commits that's why this was closed - my bad)

packages/ec-core/src/population.rs
```diff
@@ -8,11 +8,17 @@ pub trait Population {
     fn size(&self) -> usize;
 }
 
-impl<I> Population for Vec<I> {
+impl<T, I> Population for T
+where
+    // An alternative would be `T: Deref<[T]>` but this also supports
+    // hashsets, etc.
+    for<'a> &'a T: IntoIterator<Item = &'a I, IntoIter: ExactSizeIterator>,
+    T: IntoIterator<Item = I>,
```
NicMcPhee, line 16: To be honest, I'm confused here about how/why we have two different but very similar constraints on `T`. So we're saying here that `T` can be converted into an iterator over `I`s _and_ an iterator over _references_ to `I`s?

JustusFluegel: Maybe I should have commented on that but here is my reasoning:

we need the length of the population, which we can do for anything that can be turned into an `ExactSizeIter` since we can call `.len()` on that. But we don't want `T: IntoIterator<Item = I, IntoIter: ExactSizeiter>` since if we turn `T` into an iterator, we consume it, which is not what we want. Instead, we want to turn a reference into an iterator since that can't consume the underlying collection. Think of it as requiring a `.iter()` function on it.

But it also doesn't make sense to have a population which we can't turn into an iterator of individuals, so we have the second bound to require that, think of it as requiring a `.into_iterator()` method on the collection.

Basically the bounds say we need both a `.iter()` as well as a `.into_iter()` method on the collection then it is also a `Population`

JustusFluegel: In theory we could drop the second bound and make it less strict, but idk if that is what we actually want.

NicMcPhee: This makes a lot of sense – thanks! It would be good to document some of this in the code, since I think it's pretty non-obvious.

NicMcPhee: @JustusFluegel Is the plan to add some documentation explaining the two constraints?

JustusFluegel: Yes just haven't gotten to it today since I got a bit sick and pretty much stayed in bed with some tea and a book all day

NicMcPhee (approved):
Thanks again for the additional documentation/explanation.

PR #309, opened 2024-12-27
Disable unwrap lint in tests, remove manual expect's

JustusFluegel:
Removes the expects for unwrap lints all over the place and instead allows unwraps in tests using the corresponding clippy config option


NicMcPhee (approved):
This is great idea – thanks for the improvement!

JustusFluegel:
@esitsu will be happy :)

PR #312, opened 2024-12-28
Change Error and Score structs to be newtypes

### LINKED ISSUE #270: Convert `Score` and `Error` to newtypes that wrap a value
Since these only have a single named field, we don't really need the named field. It seems unlikely that we will need additional fields, so just having wrappers might be reasonable.

JustusFluegel:
Closes #270


NicMcPhee (approved):
Thanks!

PR #313, opened 2024-12-28
Change stack try_extend impl to avoid additional heap allocations 

JustusFluegel:
This also prevents deadlocks/crashes/oom errors in case of infinite or very large iterators

### DIFF
--- a/packages/push-macros/src/push_state/printing/generate_builder.rs
+++ b/packages/push-macros/src/push_state/printing/generate_builder.rs
@@ -299,7 +299,7 @@ pub fn generate_builder(
                                 ::std::iter::DoubleEndedIterator +
                                 ::std::iter::ExactSizeIterator,
                         {
-                            self.partial_state.#field.try_extend(values)?;
+                            self.partial_state.#field.push_many(values)?;
 
                             ::std::result::Result::Ok(#builder_name {
                                 partial_state: self.partial_state,
@@ -564,7 +564,7 @@ pub fn generate_builder(
                 self
                     .partial_state
                     .#exec_stack_ident
-                    .try_extend(::std::iter::IntoIterator::into_iter(program)
+                    .push_many(::std::iter::IntoIterator::into_iter(program)
                     .map(::std::convert::Into::into))?;
                 ::std::result::Result::Ok(#builder_name {
                     partial_state: self.partial_state,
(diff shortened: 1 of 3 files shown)


JustusFluegel:
Discusson from @NicMcPhee and my chat on discord:

@NicMcPhee 
> I'm a little confused in the try_extend PR, and more by the code from before than your change. There appear to be two implementations of `try_extend`, one in the `impl<A> TryExtend<A> for Stack<A>` block and one in the `impl Stack<T>` block. To be honest with you, I'm not sure why this even compiles since I can't immediately see how the compiler would distinguish between them. Based on some quick code experiments, though, it was the one in `impl Stack<T>` that was called when I created a stack and extended it, and that one hasn't changed in this PR. Do we need them both?

My Response:

> Well the one we provided in the `impl` block is a lot more restrictive since it requires `DoubleEndediterator + ExactSizeIterator` and the one from the trait is more flexible and allows us to pass Stacks to anything expecting `T: TryExtend<U>`. I bet we are using it generically somewhere and that's why we had the `TryExtend` impl as well - as for the name resolution I think it goes this way:
> 
> 1. any `impl`'s on the type directly (since you can't explicitly select them with `<T as ...>` since `<T as T>` doesn't really make sense)
> 2. if exactly one trait is in scope that provides the method use that
> 3. If multiple traits are in scope then you need to specify which one you wish to use
> 
> The impl basically changed from what we were doing before:
> 
> 1. Collect the iterator as a vector (since the iter might not implement `ExactSizeiterator` and `DoubleEndediterator`)
> 2. Create an iterator from the vector 
> 3. call the impl on the type  which requires that more restrictive constraint (which we can't express in the trait cuz the trait is not made for that) which now works since the iter from the vec supports that.
> 
>      3.1 reverse the iterator
>      3.2 check iter length + current stack is less than max stack size
>      3.3 extend the underlying vector in the stack
> 
> to what we are doing now:
> 
> 1. extend the vector as long as possible until the max stack size is reached (in the wrong order)
> 2. check that the remaining iter is emtpy (no more elements that would be appended above the max stack size
> 
>      2.1 if any remaining elements reset the stack to as it was before and return an error
> 3. reverse the appended elements in-place
> 
> This basically removes the in-between vector allocation to convert the iterator and as such only requires the allocations for the underlying stack (which should result in less allocations overall and reduce heap fragmentation from the vector that would be immediately dropped afterwards) and is overall more efficent

NicMcPhee:
I think we should probably have some of this conversation moved into the code, although I'm not 100% sure what that would be.

If nothing else, I do feel like changing the name in the `impl Stack` block to something other than `try_extend` makes sense. It would make it clearer which of these two implementations is being used, and what assumptions are being made. Does that seem reasonable?

NicMcPhee (approved):
Thanks for making those changes!

PR #314, opened 2024-12-28
Build rustdoc in ci

### LINKED ISSUE #213: Have CI check for broken `rustdoc` links
We should have the CI check for broken `rustdoc` links and fail builds if that happens.

JustusFluegel:
This enables us to check for broken intra-doc links inside our documentation.

Closes #213

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,3 +58,17 @@ jobs:
           components: clippy
           toolchain: nightly
       - run: chmod +x scripts/cargo-rustdoc-clippy && export PATH="$(pwd)/scripts/:$PATH" && cargo rustdoc-clippy -- -Dwarnings
+  rustdoc:
+    name: rustdoc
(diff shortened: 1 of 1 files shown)


JustusFluegel:
After merging this pr put the rustdoc job as a required check as well.

JustusFluegel:
We might also want to upload the docs generated here to gh-pages or something but I figured that is a task for another day.

JustusFluegel:
Oh also the clippy doctests check isn't marked as required even though we probably should do so.

NicMcPhee (approved):
Nice – thanks.

Presumably we could capture the output of this and put it somewhere like GitHub.io if we wanted to, right?

JustusFluegel:
> Presumably we could capture the output of this and put it somewhere like GitHub.io if we wanted to, right?

sure, that's what I meant with 

> We might also want to upload the docs generated here to gh-pages or something but I figured that is a task for another day.

PR #315, opened 2024-12-29
Add `Generation::into_population` method

### LINKED ISSUE #278: Add `Generation::into_population()`
At the end of a `Run`, we typically don't care about the `Generation` anymore, and only are interested in the final population. Adding an `Generation::into_population()` method would allow us to "convert" the generation into that final population without any cloning.

JustusFluegel:
Closes #278

### DIFF
--- a/packages/ec-core/src/generation.rs
+++ b/packages/ec-core/src/generation.rs
@@ -12,6 +12,10 @@ impl<P, C> Generation<P, C> {
     pub const fn population(&self) -> &P {
         &self.population
     }
+
+    pub fn into_population(self) -> P {
+        self.population
+    }
 }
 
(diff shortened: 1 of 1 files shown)


NicMcPhee (commented):
Thanks!

Would we want to implement `From<Generation> for Population` so this can happen "automagically", or are we happy with just a simple `into_population()` method here?

JustusFluegel:
> Would we want to implement `From<Generation> for Population` so this can happen "automagically", or are we happy with just a simple `into_population()` method here?

honestly I am not terribly positive that we can since `P` is generic and that might conflict with the `impl From<T> for T` in the standard library.

NicMcPhee (approved):

PR #316, opened 2024-12-29
Add `TestResults::len(&self) -> usize` method

### LINKED ISSUE #238: Create `TestResults::num_test_cases()`
I think that we call `TestResults.results.len()` in a variety of places, and having a function for that would improve readability and ergonomics.

JustusFluegel:
Also adds the `is_empty()` method on the same type since clippy complains if we have a `len()` without a `.is_empty()`

Closes #238


NicMcPhee (approved):
This looks great – thanks!

PR #317, opened 2024-12-29
Remove useless conversion to same type (nightly clippy lint fix)

JustusFluegel:
This fixes a new clippy lint that popped up in recent nightly ci runs. Together with #305 this will fix all the broken nightly clippy lints.


NicMcPhee (approved):
Thanks!

I wonder why that conversion was there in the first place? There was presumably some reason I did it, but I have no idea what that reason might be anymore.

PR #322, opened 2025-01-07
Derive `Clone` for `Score`, `Error`, and `TestResults`

NicMcPhee:
The inability to `Copy` (or at least `Clone`) various error types is turning into a substantial pain in work on [the code of the S25 EC/AI elective](https://github.com/UMM-CSci-4553-S25/course-helpers). Hopefully this will help that.

While I was here I also derived:
   - `Copy` for `Score` and `Error`
   - `Hash` while I was there since there's no reason not to and it might be helpful later
   - `Clone`, `Copy`, and `Hash` for `TestResult`, although it's not clear that we're ever going to _use_ that

I think that one thing I didn't really appreciate when a lot of this was written over a year ago (although I think that esitsu@Twitch tried to explain it to me) was that deriving, e.g., `Copy` for `Score<T>` didn't mean you could _only_ use types `T` that implemented `Copy`, but instead it means that _if_ `T` implements `Copy` _then_ `Score<T>` will also implement `Copy`. I think this confusion on my part about how trait bounds work in Rust may have led to a number of sub-optimal design decisions early on. Sighz.


JustusFluegel:
Just for future reference fyi in general there are 3 categories of macros, attribute macros, function style macros and derive macros and they each have restrictions on what they can do: Attribute macros can pretty much do anything with the expression/statement/... they are annotating, including removing or replacing parts of it, adding new attributes or statements afterwards, and so on. Pretty much the same applies for function style macros, just that they replace their invocation with some token stream and not an expression/statement.

Derive macros on the other hand get some tokens as input, the expression or statements they are annotating, but can't actually change or delete those and only add new tokens to the surrounding scope, meaning new impl blocks, etc. This also implies that a derive macro may never add bounds to the struct itsself since it can't modify its definition

JustusFluegel:
I'll try to review this tomorrow once I am at my PC again

NicMcPhee:
Thanks. Someday I'll just have to annoy esitsu@Twitch and spend some time learning more of the details about Rust macros. 😜

.vscode/settings.json
```diff
@@ -3,6 +3,7 @@
         "Bitstring",
         "Bitstrings",
         "Clojush",
+        "Cloneable",
         "Conv",
         "Dataless",
         "esitsu",
```
JustusFluegel, line 9: :joy: esitsu needed to be ignored from spellcheck xD

packages/ec-core/src/test_results.rs
```diff
@@ -349,4 +338,59 @@ mod test_results_from_vec {
...
+    }
+
+    #[test]
+    fn can_clone_test_results_of_errors_of_i32() {
+        let errors = vec![5, 8, 0, 9];
+        let test_results: TestResults<Error<i32>> = errors.into();
+        #[expect(clippy::redundant_clone, reason = "We're testing Clone")]
+        let cloned_test_results = test_results.clone();
+        assert_eq!(cloned_test_results, test_results);
+    }
```
JustusFluegel, line 358: does this really provide any value? we could just do

```rs
static_assertions::assert_impl_all!(TestResults<Error<i32>>: Clone);
```

instead if you insist but this unit test doesn't make much sense to me since we aren't actually testing any functionality/code we wrote and as such this more checks that the rust compilers derive is working correctly - which is a thing we could do I guess but I'd say we/rust has bigger issues if that breaks.

Instead we are really testing that this compiles, which the `assert_impl_all` could do just as well.

JustusFluegel: But honestly I think that isn't even necessary since removing a derive is pretty hard to do by accident, and I don't see any other reason for why this test could fail.

packages/ec-core/src/test_results.rs
```diff
@@ -349,4 +338,59 @@ mod test_results_from_vec {
...
+    //         NonCloneable(5),
+    //         NonCloneable(8),
+    //         NonCloneable(0),
+    //         NonCloneable(9),
+    //     ];
+    //     let test_results: TestResults<NonCloneable> = scores.into();
+    //     #[expect(clippy::redundant_clone, reason = "We're testing Clone")]
+    //     let cloned_test_results = test_results.clone();
+    //     assert_eq!(cloned_test_results, test_results);
+    // }
```
JustusFluegel, line 395: Again, see the above comment. If we want to test this (huge if) we could use `static_assertions::assert_not_impl_any!()` instead.

JustusFluegel (changes requested):
Can you take a look at the two comments on the tests?

packages/ec-core/src/test_results.rs
JustusFluegel: But honestly this just seems to me more like you playing around with the derive macro to see how it works, which is fine I guess if it helped you understand them better, but I wouldn't actually put it in the project and just delete it

packages/ec-core/src/test_results.rs
NicMcPhee: I agree with all this. I think I was focused on "how do I test that I added `.clone()` correctly and went overboard on the tests.

In some sense I was trying to have the tests say "These should be cloneable", in part because removing `Clone` doesn't (I think) break anything inside `ec-core`. So maybe we do want the `assert_impl` just as a way of saying explicitly that we _need_ `Clone` and aren't just including it "out of habit".

packages/ec-core/src/test_results.rs
JustusFluegel: And probably this has even less value than the above assert.

JustusFluegel (approved):
LGTM

PR #324, opened 2025-01-12
Add derive macro for Composable trait

### LINKED ISSUE #289: Add `derive(Composable)`
This would let us use the `derive` syntax instead of requiring `impl Composable for Foo {}` for basically every operator.

JustusFluegel:
Allows us to replace the manual impls all over the place with a simple derive.

Closes #289


packages/ec-macros/src/derive_composable.rs
NicMcPhee, line 1: This looks great. I don't know what "common practice" is on things like this, but it seems like it might be useful to have a comment that makes it clear that this (essentially) adds

```
impl Composable for T {}
```

for the type `T` if you `#[derive(Composable)]`?

In particular it's a little odd that all we need in the body is `{}` and I wonder if that should be noted in a comment somewhere.

(I realize that this oddness was true before you introduced the `derive`. The `derive` allows us to document that in one place instead of repeating a comment every time we implemented `Composable`, which is in every way a Good Thing IMO.)

packages/ec-macros/src/dyn_ref_impls/mod.rs
NicMcPhee, line 1: I assume you pulled this out into this file just to improve the organization of the macro code?

NicMcPhee (approved):
This looks wonderful – thanks!

I'll approve it, but feel free to add a little documentation to the macro implementation if you think that makes sense, and I can re-review it.

packages/ec-macros/src/dyn_ref_impls/mod.rs
JustusFluegel: Yes exactly and the heuristics sadly didn't catch this as a file rename with minimal changes

packages/ec-macros/src/derive_composable.rs
JustusFluegel: Sure I'll add some docs.

JustusFluegel: I added the docs but the derive macro docs stuff sadly is a bit weird so they need to be in the top level `lib.rs` - just wanted to mention that so you can find them :)

packages/ec-core/src/operator/mutator/mod.rs
```diff
@@ -174,6 +174,7 @@ pub trait Mutator<G> {
 ///
 /// // A simple `Operator` that takes a `Genome<bool>` and returns the number
 /// // of `true` values in the genome.
+/// #[derive(Composable)]
```
NicMcPhee, line 177: Good catch!

NicMcPhee (approved):

PR #325, opened 2025-01-12
Make lexicase use checked indexing

### LINKED ISSUE #241: Replace use of `[]` in `Lexicase`
Replace these unchecked accesses with checked accesses using `.get()`.

(There are probably similar uses of `[]` in other selectors; I should look at that and create issue for those as appropriate.)

JustusFluegel:
Closes #241

### DIFF
--- a/packages/ec-core/src/operator/selector/lexicase.rs
+++ b/packages/ec-core/src/operator/selector/lexicase.rs
@@ -1,5 +1,6 @@
-use std::{cmp::Ordering, mem::swap, ops::Not};
+use std::cmp::Ordering;
 
+use miette::Diagnostic;
 use rand::{Rng, prelude::SliceRandom};
 
 use super::{Selector, error::EmptyPopulation};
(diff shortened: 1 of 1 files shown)


packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -54,22 +74,35 @@ where
...
             winners.clear();
-            winners.push(candidates[0]);
-            let mut current_best_result = &winners[0].test_results().results[test_case_index];
+            winners.push(initial_winner);
+
+            let mut current_best_result = initial_winner
+                .test_results()
+                .results
+                .get(test_case_index)
+                .ok_or(LexicaseError::MissingTestCase {
```
NicMcPhee, line 93: Should this be `ok_or_else()`? I realize that there's not any "interesting" computation here, but the `MissingTestCase` struct is getting rebuilt every time through the loop, and will almost never actually get used.

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -54,22 +74,35 @@ where
...
+                    total_cases: self.num_test_cases,
+                    current_index: test_case_index,
+                })?;
+
+            for c in remaining {
+                let this_result = c.test_results().results.get(test_case_index).ok_or(
```
NicMcPhee, line 99: `ok_or_else()`?

NicMcPhee (approved):
This looks great except for the possible `ok_or_else` question. Feel free to do something about that if you think it's reasonable, otherwise I'm happy (and I'm sure you're happier as well!).

packages/ec-core/src/operator/selector/lexicase.rs
JustusFluegel: I went with `ok_or_else` at first as well but clippy wanted it this way 🤷  I'd leave it like this until we see that it makes a measurable performance impact

packages/ec-core/src/operator/selector/lexicase.rs
JustusFluegel: Same here

PR #329, opened 2025-01-19
Add readme placeholders

JustusFluegel:
This adds some basic readme placeholders since we need those to publish the crates as packages to a (local) registry. We should definitely create some more proper ones sometime soon.


NicMcPhee (approved):
Thanks!

PR #331, opened 2025-01-19
Update `rand` to `beta.3`

JustusFluegel:
This also required changing references to `rand::distr::Slice` to `rand::distr::slice::Choose`, and is a requirement to publish the local packages to a (local) registry.


NicMcPhee (approved):
Thanks!

PR #332, opened 2025-01-19
Document the publishing process

JustusFluegel:
To aid @NicMcPhee 's course registry so he can look up the process again if he gets lost.


NicMcPhee (approved):
This is super helpful – thanks a ton!

Issue #327, opened 2025-01-17
Make `WeightedPair<A,B>` struct members not `pub(crate)` anymore


JustusFluegel:
I see no reason for them beeing `pub(crate)` in the first place.

PR #333, opened 2025-01-20
Make `WeightedPair<A,B>` and `Weighted<T>` struct items private

### LINKED ISSUE #328: Make `Weighted<T>` struct members not `pub(crate)` anymore.
I see no reason for them beeing pub(crate) in the first place.

Also see #327

JustusFluegel:
All the struct items inside the two structs mentioned above were previously made `pub(crate)`. I think we accidentally introduced that during the restructuring of a non-dyn weighted selector into a more general weighted type, but never noticed that we actually don't need them to be `pub(crate)` anymore.

Making them private reduces the risk of accidental breakages through unintended direct member accesses outside the module where the structs are declared.

Closes #328 and #327

### DIFF
--- a/packages/ec-core/src/weighted/mod.rs
+++ b/packages/ec-core/src/weighted/mod.rs
@@ -11,8 +11,8 @@ pub mod with_weighted_item;
 
 #[derive(Debug)]
 pub struct Weighted<T> {
-    pub(crate) item: T,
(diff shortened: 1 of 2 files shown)


NicMcPhee (approved):
Good catch – thanks!

PR #335, opened 2025-02-02
Apply beta rustfmt

JustusFluegel:


NicMcPhee:
@JustusFluegel – I never got to this when it was new, and now it's drifted behind `main`. Should I try to bring it up to date and approve it? I actually quite like the changes, so I'd like to see it brought in.

JustusFluegel:
We should wait till next weekend for this, then beta becomes stable and our stable ci will like this more :)

NicMcPhee (approved):
Thanks for these changes. I do prefer the new formatting, so this is nice.

PR #349, opened 2025-02-22
Common derives & simple docs (part five / final part of the patchset)

JustusFluegel:
Add common derives to all applicable structs as well as provide some rudimentary documentation for all items.

This for now mostly only handles (apart from changes introduced by changes in the following crates)
- `ec-linear`
- `ec-core`
- `ec-macros`

For the sake of reviewability, `push` and `push-macros` should be done in a separate pr.

To simplify reviewability, this was divided into 5 separate that need to be processed in order because each builds on the previous ones:
- [x] #430 
- [x] #431 
- [x] #432 
- [ ] #433 
- [ ] This PR as part 5


packages/ec-core/src/individual/ec.rs
```diff
@@ -128,20 +297,40 @@ impl<GG> WithScorer for GG {
     }
 }
 
-// G is Genome
-// S is Scorer
-impl<G, D, S> Distribution<EcIndividual<G, S::Score>> for IndividualGenerator<D, S>
+impl<GenomeT, GenomeGeneratorT, ScorerT> Distribution<EcIndividual<GenomeT, ScorerT::Score>>
```
NicMcPhee, line 300: I appreciate the expansion of the generic names here, but do we need the `T` at the end of the names? It doesn't really add anything, and it's not a pattern we (I) have used anywhere else.

packages/ec-core/src/operator/composable/map.rs
```diff
@@ -64,16 +110,37 @@ where
     }
 }
 
-impl<F, Input> Operator<(Input, Input)> for Map<F>
+impl<F, Input1, Input2> Operator<(Input1, Input2)> for Map<F>
```
NicMcPhee, line 113: Nice generalization here.

packages/ec-core/src/operator/selector/tournament.rs
```diff
@@ -6,27 +6,89 @@ use rand::{Rng, prelude::IndexedRandom};
...
 impl TournamentSizeError {
+    /// Construct a new tournament size error from size of the population and
+    /// tournament
+    ///
+    /// # Safety
+    /// The caller must ensure to uphold the invarant that `tournament_size >
+    /// population_size`.
     #[must_use]
-    pub const fn new(tournament_size: NonZeroUsize, population_size: usize) -> Self {
+    pub const unsafe fn new(tournament_size: NonZeroUsize, population_size: usize) -> Self {
```
NicMcPhee, line 75: I would _really_ like to avoid using `unsafe` anywhere in the code base. I realize that I'm being pretty pedantic here, but I think it's probably better for the health of the project if we just keep that off the table, especially if we get lucky and some other people start contributing to the project.

This hardly seems like the place to be optimizing since it's onlly called if an _error_ occurs, and if that happens the program is probably about to terminate and we don't really care much about performance anymore.

packages/ec-core/src/test_results.rs
```diff
@@ -231,35 +509,313 @@ pub struct TestResults<R> {
...
+    /// ```
+    /// # use ec_core::test_results::{TestResults, Score};
+    /// let results = [Score(10), Score(20)]
+    ///     .into_iter()
+    ///     .collect::<TestResults<Score<i32>>>();
+    ///
+    /// assert_eq!(results.into_total(), Score(30));
+    /// ```
+    #[must_use]
+    pub fn into_total(self) -> R {
```
NicMcPhee, line 580: Parts of this PR arguably got a little out of control. These additions to `TestResults` make a lot of sense, but it really would have made more sense if they were in a separate, more focused PR.

NicMcPhee (dismissed):
Thanks for all the work on this, and sorry that I've been slow to respond to it.

The big thing that I would _really_ like to see unwound is the introduction of `unsafe`.

One thing that I also noticed is that this adds derives for `Ord` and `PartialOrd` to a _bunch_ of types (like generators and operators) where it's not obvious what an meaningful ordering would look like for those types, or why we would want to order them. Does this come from trying to make make some "big" type like `PushState` ordered? I'm not sure we'd want to have that derive everywhere like that. I know that we've had conversations about whether `PushState` should be `Eq` and/or `Ord`, but I've honestly lost the context on those and don't remember what the issues were. Spreading `Ord` all over the world, though, doesn't seem great?

packages/ec-core/src/individual/ec.rs
JustusFluegel: The `T` is really to differentiate the generics from the Traits and ensure no name collisions - my feeling was since you felt it was necessary to add comments for what the generics are just name them as such :)

JustusFluegel:
> One thing that I also noticed is that this adds derives for `Ord` and `PartialOrd` to a _bunch_ of types (like generators and operators) where it's not obvious what an meaningful ordering would look like for those types, or why we would want to order them. Does this come from trying to make make some "big" type like `PushState` ordered? I'm not sure we'd want to have that derive everywhere like that. I know that we've had conversations about whether `PushState` should be `Eq` and/or `Ord`, but I've honestly lost the context on those and don't remember what the issues were. Spreading `Ord` all over the world, though, doesn't seem great?

No, this really stems from trying to follow https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits from the rust api-guidelines. In particular Ord and PartialOrd make sense, since they allow the usage of these Types in BTreeSet's and BTreeMaps, and it's always frustrating if you want to use types as such but they don't implement the traits.

packages/ec-core/src/operator/selector/tournament.rs
JustusFluegel: Well the unsafe here is not really necessary from a soundness / ub side of view, I really only added it to highlight to the user that when constructing the error that tournament_size > population_size is satisfied, since otherwise the error makes zero sense. As such it's not "unsafe" in the classical sense of allowing raw pointer derefs/... but more of a marker of "hey be careful when constructing this" - but if you don't like that we can just as easily not mark it as unsafe

JustusFluegel: (and still keep the debug_assert)

packages/ec-core/src/test_results.rs
JustusFluegel: Fair enough - I think in the process of adding examples and docs I found a few places where apis were like partially implemented (like having a .total() but not a .into_total()) and just went ahead and fully implemented them. I tried not to add too much fully new functionality tho ;)

JustusFluegel:
On another note this PR is still not fully finished, a bunch of types aren't documented yet. That doesn't mean that we shouldn't do other work in the meantime that might break this, it's my responsibility to keep this here up to date, but it does mean that we shouldn't merge this just yet.

packages/ec-core/src/operator/selector/tournament.rs
JustusFluegel: Also I am pretty against a strict "no unsafe" policy in this repo, just from the standpoint of allowing future performance improvements here. One focus of the project apart from making this pretty usable is to make one of the fastest ec implementations currently available and as such keeping that door open seems like a good idea to me.

What I would maybe consider instead is adding a #[warn(unsafe_code)] or a #[error(unsafe_code)] so that we need to explicitly silence the lint with a #[expect()] and  provide a reason. We could then even add a github bot which scans the diff for new unsafe additions and adds a comment to highlight it in review, including the reason for it beeing there (from the expect)

JustusFluegel: I always found strict "no unsafe" policies a bit overblown, I think unsafe is good in the sense of that it highlights areas that should be reviewed more closely, but not allowing unsafe code at all really limits some of the more low level stuff you can do, also for performance optimizations, and has little to no benefit, especially here.

What is the goal of a  "no unsafe" policy in the first place? It's probably to ensure soundness and prevent undefined behavior and as such prevent crashes and security bugs, but having well-reviewed unsafe code with as small a scope as possible (not just marking everything unsafe but really only the small specific places where it's needed, with safety comments explaining why invariants are kept) really helps in that regard and also highlights it during review, meaning undefined behaviour/security issues are pretty rarely introduced by that.

Also, for unhindered-ec specifically: What is the attack surface that that needs to be a strict requirement? If this crashes for some reason or has soundness issues, yes that would be really annoying to debug but I see no real harm that can be done that way.

And not to sound like I am advocating to just always use unsafe or "just use c", I only mean small, well-reviewed pieces of unsafe code can have huge benefits, and abstracting that away behind safe apis is one of the reasons rust is as powerful a programming language as it is, the safety invariants it provides are really only possible because of encapsulated unsafe code "under the hood". 

tl;dr: I find it pretty overblown to "fear" unsafe code, it can be a good thing as long as it's used properly.

JustusFluegel: Also: Not to say that this specific use of unsafe is a good example of that, if you prefer this just be a safe function we can easily do that :). I think it's just a place where the function signature doesn't really express the invariants of the type and as such marking the function as unsafe is a good way to alert the consumer of that function of that :)

JustusFluegel: Also to clear up one misconception: marking a function as unsafe isn't the same as using a unsafe block, at least since edition 2024. A unsafe function doesn't allow execution of unsafe code within it without an additional unsafe block around that. It only means a unsafe context is required to call this function

packages/ec-core/src/distributions/collection.rs
```diff
@@ -1,92 +1,194 @@
...
+///
+/// let mut rng = rng();
+///
+/// let single_element: i32 = singular_distribution.sample(&mut rng);
+/// # let _ = single_element;
+/// let collection: Vec<i32> = collection_distribution.sample(&mut rng);
+/// assert_eq!(collection.len(), 10)
+/// ```
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Default)]
+pub struct Collection<C> {
```
NicMcPhee: Either move the `Generator` -> `Collection` rename into its own PR or join it with the `Distribution` renamings.

Cargo.toml
```diff
@@ -13,6 +13,7 @@ documentation = "https://github.com/unhindered-ec/unhindered-ec"
 repository = "https://github.com/unhindered-ec/unhindered-ec.git"
 edition = "2021"
 license = "MIT OR Apache-2.0"
+publish = false
```
NicMcPhee: Should this be it's own PR?

NicMcPhee (changes requested):
There is a lot of wonderful stuff, but it would be nice to split some of it out (see review comments) into new PRs. Thanks!

packages/ec-core/src/operator/composable/map.rs
```diff
@@ -27,53 +60,102 @@ impl<T> std::fmt::Display for MapError<T> {
...
+        // Option<T> instead, since then no additional tag information needs to be
+        // stored.
+        let mut output: [Option<F::Output>; N] = [const { None }; N];
+
+        for (index, (input, output)) in input.into_iter().zip(output.as_mut_slice()).enumerate() {
+            output.replace(self.f.apply(input, rng).map_err(|e| MapError(e, index))?);
+        }
+
+        Ok(output.map(Option::unwrap))
     }
```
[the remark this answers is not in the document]
NicMcPhee: As far as I can tell, this is just wrong, and that `[T; N]::map` doesn't require that `F::Output` implement `Copy`.  [The documentation](https://doc.rust-lang.org/stable/std/primitive.array.html#impl-%5BT;+N%5D) doesn't suggest any need for `Copy` on the function output. Maybe it had such a requirement in the past, but it doesnt' seem to have it now. @JustusFluegel: Am I reading this correctly?

packages/ec-core/src/distributions/finite.rs
```diff
@@ -0,0 +1,79 @@
...
+/// use ec_core::distributions::finite::Finite;
+/// use rand::distr::slice::Choose;
+///
+/// assert_eq!(
+///     Finite::sample_space_size(&Choose::new(&[1, 2, 3]).unwrap()),
+///     NonZero::new(3).unwrap()
+/// );
+/// ```
+pub trait Finite {
+    /// The size of this distributions sample space.
```
NicMcPhee:
```suggestion
    /// The size of this distribution's sample space.
```

packages/ec-core/src/distributions/finite.rs
```diff
@@ -0,0 +1,79 @@
+use std::num::NonZeroUsize;
+
+use rand::distr::slice::Choose;
+
+/// A distribution which is finite, i.e. has a finite sample space
+///
+/// This can be useful to for example determine the probability of elements in a
+/// uniform distribution.
```
NicMcPhee:
```suggestion
/// This can be useful to, for example, in determining 
/// the probability of choosing any particular element
/// from a uniform distribution.
```

packages/ec-core/src/distributions/wrappers/choose_cloning.rs
```diff
@@ -1,33 +1,77 @@
...
+/// alternative that takes ownership of the collection instead.
+///
+///
+/// # Example
+/// ```
+/// # use rand::{rng, distr::{Distribution, slice::Empty}};
+/// # use ec_core::distributions::wrappers::choose_cloning::ChooseCloning;
+/// #
+/// # fn main() -> Result<(), Empty> {
+/// let distribution = ChooseCloning::new(&[1, 2, 3])?;
```
NicMcPhee: I wouldn't want to block the PR for this, but would it make sense to have the example use types where cloning might matter (like `String`) instead of using a type that also implements `Copy`?

This also applies to some of the other examples below.

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
```diff
@@ -1,17 +1,36 @@
...
+/// let my_collection = [0, 1, 2, 3];
+/// let my_distribution = ChooseCloningOwning::new(my_collection)?;
+///
+/// let my_sample = my_distribution.sample(&mut rng());
+/// # let _ = my_sample;
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub struct ChooseCloningOwning<T, U> {
```
NicMcPhee: The documentation and example don't say anything about the generic `U`, i.e., what role it plays or why it's there. I think we should say something about it rather than just have the semi-mysterious `_p: PhantomData<U>` line.

We may want to include mentions/explanations of `U` in some of the subsequent documentation as well (like for `new()`).

packages/ec-core/src/distributions/collection.rs
```diff
@@ -1,92 +1,194 @@
...
+/// - [`ConvertToCollectionDistribution::to_collection`], a borrowing
+///   constructor, borrowing the child distribution.
+///
+/// Usage:
+///
+/// ```
+/// # use rand::distr::StandardUniform;
+/// # use ec_core::distributions::collection::{Collection, ConvertToCollectionDistribution};
+/// #
+/// let collection_distribution: Collection<StandardUniform> = StandardUniform.into_collection(10);
```
NicMcPhee: Not a huge deal, but it would be nice if it were clearer than `10` is the size/length of the generated/sampled collection. Maybe just give it a name?

```rust
let collection_size = 10;
/// ...
let collection_distribution: Collection<StandardUniformL> 
    = StandardUniform.into_collection(collection_size);
```

packages/ec-core/src/distributions/collection.rs
```diff
@@ -1,92 +1,194 @@
...
-    /// Convert the type into a `CollectionGenerator` that generates collections
-    /// of the specified size, using `self` to generate the individual elements.
-    /// This takes ownership of the type.
-    fn into_collection_generator(self, size: usize) -> Generator<Self>
+/// Helper methods to apply the [`Collection`] [`Distribution`] adapter to other
+/// [`Distribution`]'s.
+///
+/// This trait isn't meant to be implemented externally, rather
+/// it is an extension trait that is blanket-implemented on all
+/// types, although it is most useful for [`Distribution`]'s.
```
NicMcPhee: Presumably if you use one of these on something that isn't a `Distribution` there will typically be an error somewhere as a result. What would that error look like? Is this a place where we'd get one of those obscure Rust trait errors that would be hard to figure out?

packages/ec-core/src/distributions/collection.rs
```diff
@@ -1,92 +1,194 @@
...
-impl<T, C> Distribution<Vec<T>> for Generator<C>
+impl<T, C> Distribution<Vec<T>> for Collection<C>
 where
     C: Distribution<T>,
 {
+    /// Sample the [`Vec`] collection from this [`Collection`] distribution, of
+    /// length `length` as specified by the distribution.
     fn sample<R: rand::Rng + ?Sized>(&self, rng: &mut R) -> Vec<T> {
-        (&self.element_generator)
+        (&self.item_distribution)
```
NicMcPhee: All three of these implementations are identical. They're simple, so it's not a huge issue, but it does trigger my code duplication gnome.

I spent longer than I probably should have looking for an "easy" way to avoid this, and everything complicates things in various ways that make me somewhat inclined to leave things as they are, TBH.

packages/ec-core/src/distributions/conversion.rs
```diff
@@ -120,44 +283,54 @@ where
...
-}
-
 impl<'a, T> IntoDistribution<&'a T> for &'a [T] {
     type Distribution = Choose<'a, T>;
 
-    type Error = EmptySlice;
+    type Error = Empty;
 
+    /// Creates a [`Distribution`] which uniformly samples a single element from
+    /// a [`&[U]`](primitive@slice) and returns a reference to it.
```
NicMcPhee: Is there a typo here?

```suggestion
    /// a [`&[T]`](primitive@slice) and returns a reference to it.
```

packages/ec-core/src/individual/mod.rs
```diff
@@ -1,11 +1,130 @@
...
+///     fn genome(&self) -> &Bitstring {
+///         &self.genome
+///     }
+///
+///     fn test_results(&self) -> &Score<u32> {
+///         &self.score
+///     }
+/// }
+/// ```
+/// [^ec-linear-usage]: Note that this example uses [`ec-linear`][ec-linear] which is not a
```
NicMcPhee:
```suggestion
/// [^ec-linear-usage]: Note that this example uses the `BitString` type from [`ec-linear`][ec-linear] which is not a
```

NicMcPhee (commented):
This isn't done, but it gets some feedback out there to work on.

Thanks a ton for all this work – it definitely makes the world a better place!

packages/ec-core/src/distributions/finite.rs
JustusFluegel: I did 

```rs
/// This can be useful, for example, when determining 
/// the probability of choosing any particular element
/// from a uniform distribution.
```

since your suggestion sounded kinda wrong to me (to and then in)

packages/ec-core/src/distributions/collection.rs
JustusFluegel: I don't think it's a huge  issue and the implementation is after all made to abstract over the different types in other places so I'd be inclined to just leave it.

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
JustusFluegel: I don't know if being explicit everywhere is really necessary since both generics can almost always be inferred, but I added a section on the generics to the main struct's docs.

packages/ec-core/src/individual/ec.rs
```diff
@@ -79,37 +158,130 @@ impl<G: Display, R: Display> Display for EcIndividual<G, R> {
...
+/// // StandardUniform is a `rand` distribution allowing
+/// // us to generate i32's uniformly as genomes for this example
+/// let genome_distribution = StandardUniform;
+/// let scorer = FnScorer(|x: &i32| 100_i32.abs_diff(*x));
+///
+/// let individual_distribution = IndividualDistribution::new(genome_distribution, scorer);
+///
+/// let my_individual: EcIndividual<_, _> = individual_distribution.sample(&mut rand::rng());
+/// #
+/// # assert!((i32::MIN..=i32::MAX).contains(my_individual.genome()));
```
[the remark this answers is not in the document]
JustusFluegel: well they really don't do anything fancy since we really check that it is a i32 value which all `i32`'s are so there is an argument for them to just be removed :) So i don't think they provide much value other than making clippy happy about the usage of `my_individual` here.

packages/ec-core/src/distributions/collection.rs
NicMcPhee: We're going to leave alone for now, but created #434 to explore these design questions.

packages/ec-core/src/individual/mod.rs
```diff
@@ -1,11 +1,130 @@
...
+///     dependency of this package to demonstrate some concepts which need
+///     concrete implementations. If you want to replicate this example, make
+///     sure [`ec-linear`][ec-linear] is installed.
+///
+/// [ec-linear]: #
```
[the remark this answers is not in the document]
NicMcPhee: Thanks for the explanation. I think that creating the issue sounds like a good plan.

Issue #350, opened 2025-02-23
Implement "common" instructions that are shared across types


NicMcPhee:
There are a host of instructions that are shared across all stack types, which we should implement. 

Clojush has the following "standard" things that Clojush provided for all stack types:

- [ ] `pop` (which we have)
- [ ] `dup`
- [ ] `dup-times` (creating `n` copies of the top item of a stack)
- [ ] `dup-items` (duplicate the top `n` items on a stack, once each, preserving order)
- [ ] `swap` (swaps the top two items of a stack)
- [ ] `rot` (rotates the top 3 items of a stack, bringing the 3rd to the top and pushing the 1st and 2nd down one)
- [ ] `flush` (remove all the items from a stack)
- [ ] `eq` (compares the top two items on a stack, pushing the result on the bool stack)
- [ ] `stack-depth` (pushes the depth of this stack onto the int stack)
- [ ] `yank` (yanks the nth item, where `n` is from the int stack, of the stack up to the top; take the bottom item if `n` is larger than the size of the stack)
- [ ] `yank-dup` (`yank`, but copies instead of moves)
- [ ] `deep-dup` (same as `yank-dup`, but counts from the bottom of the stack instead of the top, duplicating the top item if `n` is larger than the stack size)
- [ ] `shove` (shoves the top item down into the stack, using the top of the int stack, `n`, to indicate the target position; if `n` exceeds the stack size, then place at the bottom of the stack)
   - Why isn't there a `shove-dup`?
- [ ] `empty?` (puts `true` on the bool stack if this stack is empty, `false` otherwise)

For many of these I think we first want to make sure there are associated function in the `Stack` type, and then the instruction just calls that command to do the work. 

I don't want to need to complete all of these in order to clear this, so I'm going to make this a tracking issue. For starters I would prioritize:

Some basic stack actions:

- `dup`
- `stack-depth`
- `empty?`

The basic comparison operators:

- `eq`
- `ne`
- `le`
- `leq`
- `gt`
- `gte`

NicMcPhee:
#359 gives us the comparison tests, so that part is now done.

PR #351, opened 2025-02-23
Add common/shared stack instructions

NicMcPhee:
This is a step towards completing #350, implementing a host of common instructions across all the stack types. Here I'm setting the infrastructure for `CommonInstruction` and implementing three of the most frequently used stack manipulation instructions:

- [x] `push_value`
- [x] `dup`
- [x] `stack_depth`
- [x] `is_empty`
- [x] `pop` (as a Push instruction)
- [x] `swap`
- [x] `flush`

I was initially going to also do `eq` here, but @JustusFluegel pointed out that the quality and comparison tests raise some interesting questions/possibilities (see notes on #350) so I think that should be a separate PR.

---

> [!NOTE]
> Below is the summary of these changes generated by Copilot when I asked it to do a code review. I think are pretty helpful, actually, so I copied them up here.

This PR lays the groundwork for a shared instruction infrastructure and implements a series of common stack instructions (pop, dup, is_empty, flush, stack_depth, swap, push_value) for the push virtual machine. Key changes include the addition of a new common module with const constructor functions for these instructions, updates to the float and integer instruction enums to use these new constructors, and corresponding adjustments in examples, benchmarks, and tests.

File | Description
-- | --
packages/push/src/instruction/common/pop.rs, dup.rs, is_empty.rs, flush.rs, stack_depth.rs, swap.rs, push_value.rs | New common instruction implementations; added const constructors and integration tests.
packages/push/src/instruction/float.rs and int/mod.rs | Updated instruction enum variants to leverage the new shared instruction constructors.
packages/push/src/instruction/exec/dup_block.rs | Minor change in error mapping on duplicated block handling.
packages/push/examples/complex_regression/main.rs, benches/regression-evaluation.rs, and other modules | Updated usage of instruction constructors to use new API (e.g. using FloatInstruction::dup() in place of a raw variant).


packages/push/src/instruction/common/dup.rs
```diff
@@ -0,0 +1,184 @@
...
+///    - Stack of type `T`
+///      - One value
+///
+/// # Behavior
+///
+/// The `Dup` instruction clones the top block on the stack of type `T`,
+/// leaving both the original and the copy on that stack.
+///
+/// If stack is empty, this is a no-op,
+/// and the state is returned unchanged.
```
JustusFluegel, line 24: We actually return a recoverable `Underflow` error so idk how accurate this comment is.

packages/push/src/instruction/common/flush.rs
```diff
@@ -0,0 +1,93 @@
...
+/// An instruction that flushes (i.e., empties) a stack of type `T`.
+///
+/// # Inputs
+///
+/// The `Flush` instruction takes no inputs from the stacks.
+///
+/// # Behavior
+///
+/// The `Flush` instruction removes all the elements from the stack of type `T`.
+///
```
JustusFluegel, line 17: Might be worth it to explicitly note what happens with an already empty stack?

packages/push/src/instruction/common/flush.rs
```diff
@@ -0,0 +1,93 @@
...
+/// | ------------- | ------------- | ------------- |
+/// | empty | ✅ | The stack is unchanged |
+/// | not empty | ✅ | All the elements are removed from the stack |
+///
+/// # Errors
+///
+/// If the stack access returns any error other than a
+/// [`StackError::Underflow`](crate::push_vm::stack::StackError::Underflow)
+/// then this returns that as a [`Error::Fatal`](crate::error::Error::Fatal)
+/// error.
```
JustusFluegel, line 46: Does this section really make sense in the context that this instruction can't fail?

packages/push/src/instruction/common/flush.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+
+impl<S, T> Instruction<S> for Flush<T>
+where
+    S: HasStack<T>,
+{
+    type Error = PushInstructionError;
+
+    fn perform(&self, mut state: S) -> crate::error::InstructionResult<S, Self::Error> {
```
JustusFluegel, line 64: In the other instructions we just import `InstructionResult`

packages/push/src/instruction/common/flush.rs
```diff
@@ -0,0 +1,93 @@
...
+            .with_max_stack_size(3)
+            .with_int_values([1, 2, 3])
+            .unwrap()
+            .with_no_program()
+            .build();
+        let flush = Flush::<i64>::new();
+        let state = flush.perform(state).unwrap();
+        assert_eq!(state.stack::<i64>().size(), 0);
+    }
+}
```
JustusFluegel, line 93: Maybe add a test for flushing an already empty stack?

packages/push/src/instruction/common/is_empty.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+impl<S, T> Instruction<S> for IsEmpty<T>
+where
+    S: HasStack<T> + HasStack<bool>,
+{
+    type Error = PushInstructionError;
+
+    fn perform(&self, state: S) -> crate::error::InstructionResult<S, Self::Error> {
```
JustusFluegel, line 75: same here, we should just `use ...::InstructionResult`

packages/push/src/instruction/common/pop.rs
```diff
@@ -0,0 +1,140 @@
...
+            .with_no_program()
+            .build();
+        let pop = Pop::<i64>::default();
+        let result = pop.perform(state).unwrap();
+        assert_eq!(result.stack::<i64>().size(), 0);
+    }
+
+    // Test that `Pop` works on stacks of 1..100 elements of type `i64`.
+    #[proptest]
+    fn pop_stacks_with_multiple_items(#[strategy(vec(any::<i64>(), 1..100))] values: Vec<i64>) {
```
JustusFluegel, line 119: `#[any(size_range(1..100).lift())]` would probably be more concise

packages/push/src/instruction/common/pop.rs
```diff
@@ -0,0 +1,140 @@
...
+            .unwrap()
+            .with_no_program()
+            .build();
+        let pop = Pop::<i64>::default();
+        let result = pop.perform(state).unwrap();
+        #[expect(
+            clippy::arithmetic_side_effects,
+            reason = "The length of `values` is less than 100, so adding 1 is guaranteed to be \
+                      safe."
+        )]
```
JustusFluegel, line 132: this safety comment doesn't make any sense in the context of the arithmetic operation we actually perform

packages/push/src/instruction/common/push_value.rs
```diff
@@ -0,0 +1,125 @@
...
+/// | ------------- | ------------- | ------------- |
+/// | false | ✅ | The internal value is pushed onto the stack |
+/// | true | [‼️..](crate::push_vm::stack::StackError::Overflow) | Program is terminated |
+///
+/// # Errors
+///
+/// If the stack access returns any error other than a
+/// [`StackError::Underflow`](crate::push_vm::stack::StackError::Underflow)
+/// then this returns that as a [`Error::Fatal`](crate::error::Error::Fatal)
+/// error.
```
JustusFluegel, line 49: Same as above idk how much sense this comment makes given the table above

packages/push/src/instruction/common/push_value.rs
```diff
@@ -0,0 +1,125 @@
...
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
+pub struct PushValue<T>(pub T);
```
JustusFluegel, line 51: I would add a new just for consistency with all the other instructions (keep this field pub tho)

packages/push/src/instruction/common/push_value.rs
```diff
@@ -0,0 +1,125 @@
...
+
+impl<S, T> Instruction<S> for PushValue<T>
+where
+    S: Clone + HasStack<T>,
+    T: Clone,
+{
+    type Error = PushInstructionError;
+
+    fn perform(&self, state: S) -> InstructionResult<S, Self::Error> {
+        state.with_push(self.0.clone()).map_err_into()
```
JustusFluegel, line 61: Not really a comment for this specific pr but one question I have is if it would make sense that the perform on the instruction actually consumes the instruction? I don't quite remember why we choose `&self` over just `self` initially or if there is a reason we did so/a pattern this enables.

Because if this took just `self` then there would for example be no need for cloning here.

packages/push/src/instruction/common/push_value.rs
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+impl<T> Display for PushValue<T>
+where
+    T: Display,
+{
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Push({})", self.0)
+    }
```
JustusFluegel, line 71: Are we using [`derive_more`](https://docs.rs/derive_more`) in this project already? if we do just use the derive if we don't I'd keep this as is.

packages/push/src/instruction/common/stack_depth.rs
```diff
@@ -0,0 +1,194 @@
...
+///
+/// If the int stack is full before this instruction is performed, then
+/// a fatal [`StackError::Overflow`](crate::push_vm::stack::StackError::Overflow) is returned,
+/// terminating the running of the program.
+///
+/// If the size of the stack exceeds the maximum value of an
+/// `i64`, then `i64::MAX` will be pushed onto the int stack. This would only
+/// happen if there are 2^63 items on `Stack<T>`, which probably should
+/// not be allowed to happen since that is almost certainly the result
+/// of an infinite loop or similar and we should use `.with_max_stack_size()`
```
JustusFluegel, line 38: might be worth to document on `.with_max_stack_size` that this instruction can "soft"-break if we set it too high.

packages/push/src/instruction/common/stack_depth.rs
```diff
@@ -0,0 +1,194 @@
...
+///
+/// Note that if the size of `Stack<T>` is too big to fit in an `i64`, then
+/// `i64::MAX` will be pushed onto the int stack instead.
+///
+/// # Errors
+///
+/// If the stack access returns any error other than a
+/// [`StackError::Underflow`](crate::push_vm::stack::StackError::Underflow)
+/// then this returns that as a [`Error::Fatal`](crate::error::Error::Fatal)
+/// error.
```
JustusFluegel, line 68: Same as above, idk how much sense this comment makes given the above table

packages/push/src/instruction/common/swap.rs
```diff
@@ -0,0 +1,164 @@
...
+///       - ✅: success (so two copies of the top block on the `Exec stack`)
+///       - ❗: recoverable error, with links to the error kind
+///       - ‼️: fatal error, with links to the error kind
+///    - The "Note" column briefly summarizes the action state in that case
+///
+/// | `Stack<T>`  | Success | Note |
+/// | ------------- | ------------- | ------------- |
+/// | exists, exists  | ✅ | The top two values are swapped |
+/// | exists, missing | [❗…](crate::push_vm::stack::StackError::Underflow) | State is unchanged |
+/// | missing, irrelevant | [❗…](crate::push_vm::stack::StackError::Underflow) | State is unchanged |
```
JustusFluegel, line 42: I'd probably rather create two columns which indicate each value rather than separating them with a comma inside one

packages/push/src/instruction/common/swap.rs
```diff
@@ -0,0 +1,164 @@
...
+///
+/// # Errors
+///
+/// If the stack access returns any error other than a
+/// [`StackError::Underflow`](crate::push_vm::stack::StackError::Underflow)
+/// then this returns that as a [`Error::Fatal`]
+/// error.
```
JustusFluegel, line 49: Same as above, idk how much sense this comment makes given the above table.

packages/push/src/instruction/common/swap.rs
```diff
@@ -0,0 +1,164 @@
...
+            result.error(),
+            &PushInstructionError::StackError(StackError::Underflow {
+                num_requested: 2,
+                num_present: 0
+            })
+        );
+    }
+
+    #[test]
+    fn swap_singleton_stack() {
```
JustusFluegel, line 109: is singleton really the right word for this? For me this has a very specific meaning in programming so I'd rather have something like `swap_single_value_stack() { }` or something akin to it.

packages/push/src/instruction/exec/dup_block.rs
```diff
@@ -75,7 +75,7 @@ where
         state
             .stack::<PushProgram>()
             .top()
-            .map_err(PushInstructionError::from)
+            // .map_err(PushInstructionError::from)
```
JustusFluegel, line 78: does this have a reason for still beeing here at all if it's not needed?

packages/push/src/instruction/common/swap.rs
```diff
@@ -0,0 +1,164 @@
...
+        let result = swap.perform(state).unwrap();
+        assert_eq!(result.stack::<i64>().size(), values.len());
+        // We need to swap the first two elements of the set of values
+        // to mimic the swap, and then reverse the set of values since
+        // they are essentially reversed when they are initially pushed
+        // onto the stack.
+        values.swap(0, 1);
+        values.reverse();
+        // Now the set of values of the stack should be the same as `values`.
+        assert_eq!(result.stack::<i64>(), &values);
```
JustusFluegel, line 162: we should probably move to this style of assert on stacks more often, instead of using e.g. `.size()` followed by `.top2()`

packages/push/src/instruction/float.rs
```diff
@@ -13,8 +19,15 @@ use crate::{
...
-    #[strum(to_string = "Push({0})")]
-    Push(OrderedFloat<f64>),
+    Pop(Pop<OrderedFloat<f64>>),
+    #[strum(to_string = "{0}")]
+    Push(PushValue<OrderedFloat<f64>>),
+    Dup(Dup<OrderedFloat<f64>>),
+    Swap(Swap<OrderedFloat<f64>>),
+    IsEmpty(IsEmpty<OrderedFloat<f64>>),
+    StackDepth(StackDepth<OrderedFloat<f64>>),
+    Flush(Flush<OrderedFloat<f64>>),
```
JustusFluegel, line 29: As I've mentioned on stream before we should probably implement `From<>` for all of these.

packages/push/src/instruction/int/clamp.rs
```diff
@@ -0,0 +1,131 @@
...
+/// | Greater than Max | exists | exists | Max | ✅ | Value is replaced by Max |
+/// | In range [Min, Max] | exists | exists | Value | ✅ | Value remains unchanged |
+/// | missing | irrelevant | irrelevant | irrelevant| [❗…](crate::push_vm::stack::StackError::Underflow) | State is unchanged |
+///
+/// # Errors
+///
+/// If the stack access returns any error other than a
+/// [`StackError::Underflow`](crate::push_vm::stack::StackError::Underflow)
+/// then this returns that as a [`Error::Fatal`](crate::error::Error::Fatal)
+/// error.
```
JustusFluegel, line 57: Same thing as everywhere else.

packages/push/src/instruction/int/clamp.rs
```diff
@@ -0,0 +1,131 @@
...
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
+pub struct Clamp;
```
JustusFluegel, line 59: We should make this a `Clamp<T: Ord>{ _p : PhantomData<T> }` and then just use `Ord:.clamp`

packages/push/src/instruction/int/clamp.rs
```diff
@@ -0,0 +1,131 @@
...
+            .top3()
+            .map_err(PushInstructionError::from)
+            .map(|(&value, &min, &max)| {
+                if value < min {
+                    min
+                } else if value > max {
+                    max
+                } else {
+                    value
+                }
```
JustusFluegel, line 79: Even if this is not generic just use `Ord::clamp` here :)

packages/push/src/instruction/int/clamp.rs
```diff
@@ -0,0 +1,131 @@
...
+            .with_max_stack_size(3)
+            .with_int_values([15, 10, 20])
+            .unwrap()
+            .with_no_program()
+            .build();
+        let result = Clamp.perform(state).unwrap();
+        assert_eq!(result.stack::<i64>().top().unwrap(), &15);
+        assert_eq!(result.stack::<i64>().size(), 1);
+    }
+}
```
JustusFluegel, line 131: Still needs a test for the error case.

packages/push/src/instruction/int/mod.rs
```diff
@@ -1,26 +1,41 @@
...
-    #[strum(to_string = "Push({0})")]
-    Push(i64),
+    Pop(Pop<i64>),
+    #[strum(to_string = "{0}")]
+    Push(PushValue<i64>),
+    Dup(Dup<i64>),
+    Swap(Swap<i64>),
+    IsEmpty(IsEmpty<i64>),
+    StackDepth(StackDepth<i64>),
+    Flush(Flush<i64>),
```
JustusFluegel, line 33: Same as with the other group, `From` impls would be nice.

packages/push/src/instruction/mod.rs
```diff
@@ -75,12 +76,12 @@ impl PushInstruction {
...
     pub fn push_int(i: i64) -> Self {
-        IntInstruction::Push(i).into()
+        IntInstruction::push(i).into()
     }
 
     #[must_use]
     pub fn push_float(f: OrderedFloat<f64>) -> Self {
-        FloatInstruction::Push(f).into()
+        FloatInstruction::push_ordered_float(f).into()
     }
```
JustusFluegel, line 85: Does it make sense to just align these and do push_ordered_float and push_float?

packages/push/src/push_vm/stack.rs
```diff
@@ -312,6 +314,51 @@ impl<T> Stack<T> {
...
+    /// of the triple will be the top element of the stack, the second element
+    /// of the triple will be the second-to-top element of the stack, and the
+    /// third element of the triple will be the third-to-top element of the
+    /// stack.
+    ///
+    /// # Errors
+    ///
+    /// Returns `StackError::Underflow` if the stack has less than three
+    /// elements.
+    pub fn top3(&self) -> Result<(&T, &T, &T), StackError> {
```
JustusFluegel, line 328: I'd approve this for this pr but we should really figure out a way to make this more general like my proposed `top_n` with a potential rework.

packages/push/src/push_vm/stack.rs
```diff
@@ -597,4 +673,61 @@ mod test {
...
+        stack.push(false).unwrap();
+        let result = stack.pop3().unwrap_err();
+        assert_eq!(
+            result,
+            StackError::Underflow {
+                num_requested: 3,
+                num_present: 2
+            }
+        );
+    }
```
JustusFluegel, line 732: Also add the above tests for  `pop3`?

JustusFluegel (changes requested):
Take a look at the comments, mostly seems fine tho :)

packages/push/src/instruction/common/dup.rs
NicMcPhee: That's a good catch, and I think it's a actually a problem on most of the documentation. I'll fix it here, and I'll make an issue to fix it everywhere else.

packages/push/src/instruction/common/flush.rs
NicMcPhee: Nope, it doesn't. Good catch.

packages/push/src/instruction/common/push_value.rs
NicMcPhee: I think this is also a problem in lots of places. I'll mention that in the new issue as well.

packages/push/src/instruction/common/push_value.rs
NicMcPhee: OK, that's weird, and I think we don't have a very consistent set up. My first thought was that you could perform this instruction multiple times, and thus so we'd have to clone _somewhere_. But `Dup` actually clones the instruction, which would (in this case) also clone the wrapped value. So we could take ownership of the instruction and just push `self.0` on the stack.

I don't particularly care where we do the cloning, but it doesn't make sense to do it twice when once would be sufficient. I think I had been thinking in terms of multiple references to a single copy of each instruction (which is how it would essentially play out in Clojure), but given that most instructions are just labels references to them will probably be the same amount of memory as the instruction itself so there's no reason to avoid cloning instructions. It then might make some other bits simpler if `perform` could take ownership of the instruction, although I don't know of any instructions other than this one where that might matter.

packages/push/src/instruction/common/push_value.rs
NicMcPhee: A search suggests we are _not_ using `derive_more`. I know you've mentioned it a few times, but it doesn't look like we've ever brought it in.

packages/push/src/instruction/common/stack_depth.rs
NicMcPhee: I added that in `generate_builder.rs`; see what you think.

packages/push/src/instruction/common/swap.rs
NicMcPhee: In math (and at least parts of computing) "singleton" also means a set or container with exactly one element. That meaning predates the singleton pattern, which is what I'm guessing you're thinking of? I think this is an appropriate use the term here, although it does highlight the challenge of natural language as a communication system.

JustusFluegel: Yea I was thinking of the Singleton pattern, to me this name implies some kind of stack which shares it's content across instances which this obviously is not. But if you just want to go with he `singleton` name I am not gonna loose sleep over it :)

packages/push/src/instruction/common/swap.rs
JustusFluegel: But probably also not a thing we need to do in this pr

packages/push/src/instruction/int/clamp.rs
JustusFluegel: ( like `value.clamp(min,max)` )

packages/push/src/instruction/common/swap.rs
NicMcPhee: Totally agreed. I'll make another issue (#356).

packages/push/src/instruction/float.rs
NicMcPhee: Agreed. That would happen separately in each instruction type (e.g., `IntInstruction`, `FloatInstruction`, etc.), correct?

Would this be a reason to have a `Common` enum? Then we could have a single `impl From<Common<i64>> for IntInstruction` instead of having a dozen separate `From` implementations in each type. Do we think that would work?

packages/push/src/instruction/int/clamp.rs
NicMcPhee: I didn't even realize there _was_ an `Ord::clamp` – thanks!

packages/push/src/instruction/float.rs
JustusFluegel: I would, for now, even if it is a lot of copy pasta just add individual from implementations for each. Yes a common enum might help reduce the clutter but I'd say in the long term the from implementations should probably just be auto generated by some kind of common infrastructure, and a common enum brought up other issues with EnumIter that I'd just like to sidestep at least for now

packages/push/src/instruction/mod.rs
JustusFluegel: Hmm then let's just leave it for this pr.

packages/push/src/instruction/common/push_value.rs
NicMcPhee: I created an issue for this (#357) and we should come back to it at some point, maybe over the summer.

packages/push/src/instruction/common/stack_depth.rs
NicMcPhee: Sorry – those commits are now pushed.

packages/push/src/instruction/int/clamp.rs
NicMcPhee: We used `Ord::clamp` in the implementation but didn't yet make it generic. I'll do that in the equality and comparison pull request.

packages/push/src/push_vm/stack.rs
NicMcPhee: This is a good idea, but I'm going to push it off to a new issue (#358).

JustusFluegel (approved):
ack, if the ci passes let's get this merged :)

JustusFluegel:
Actually, the example run now takes ∞% more time than it did before, so I did a bisect to figure out which commit caused this. (it ended up beeing 4003583745814e6fc7c2480d99001b0077bb9645. ) I then looked at that commit and started to "bisect" inside the commit by only testing half of the changes, commenting out the other half and ended up finding that adding the `Exec::Dup` and `Exec::Push` instructions really blow up the runtime a lot, which probably makes sense to me since they can end up copying instructions and creating loops and stuff (do we already have a infinite loop detection?). I would suggest commenting them out for this pr and maybe re-add them later, or make them configurable in some way.

JustusFluegel (changes requested):
Super long runtime after adding new exec instructions

JustusFluegel (approved):
ack

PR #361, opened 2025-03-23
Add a `max_instruction_steps` check

### LINKED ISSUE #360: Add a "max instruction steps" check
We don't currently have a check for infinite (or very long) execution paths, and we really need that. It's easy enough to add to `run_to_completion()`, but we should add a `.with_max_instruction_steps()` to the build process for `PushState` that allows us to set that value when setting up a problem.

As part of doing this we should:

- [x] Uncomment out `Dup` in `ExecInstruction`
- [x] Remove the filters for `BlockDup` and `Push` in `median` and `smallest`

NicMcPhee:
Closes #360


JustusFluegel (approved):
LGTM, I also added the builder derive macro support :)

PR #365, opened 2025-04-05
Add print instructions

NicMcPhee:
This adds a simple set of printing instructions so we can actually implement several of the PSB benchmark problems.


packages/push/examples/number_io/main.rs
```diff
@@ -0,0 +1,258 @@
...
+        let f = self.1.sample(rng);
+        Input { i, f }
+    }
+}
+
+// An output for this problem is an `i64`.
+#[derive(Debug, Clone)]
+struct Output(String);
+
+// This is an implementation of the "smallest" problem from Tom Helmuth's
```
JustusFluegel, line 78: this seems wrong

packages/push/examples/number_io/main.rs
```diff
@@ -0,0 +1,258 @@
...
+// GECCO '15: Proceedings of the 17th annual conference on Genetic and
+// evolutionary computation. July 2015. ACM.
+//
+// Problem Source: C. Le Goues et al., "The ManyBugs and IntroClass
+// Benchmarks for Automated Repair of C Programs," in IEEE Transactions on
+// Software Engineering, vol. 41, no. 12, pp. 1236-1256, Dec. 1 2015.
+// doi: 10.1109/TSE.2015.2454513
+//
+// This problem is quite easy if you have a `Min` instruction, but can be
+// more a bit more difficult without that instruction.
```
JustusFluegel, line 91: same here

packages/push/examples/number_io/main.rs
```diff
@@ -0,0 +1,258 @@
...
+    let Ok(output) = final_state.stdout_string() else {
+        return penalty_value;
+    };
+    damerau_levenshtein(&output, expected)
+}
+
+fn instructions() -> impl Iterator<Item = PushInstruction> {
+    let int_instructions = IntInstruction::iter()
+        // Restore this line to remove `Min` from the instruction set.
+        // .filter(|&i| i != IntInstruction::Min)
```
JustusFluegel, line 248: is this relevant for this problem?

packages/push/src/instruction/float.rs
```diff
@@ -28,7 +30,10 @@ pub enum FloatInstruction {
     IsEmpty(IsEmpty<OrderedFloat<f64>>),
     StackDepth(StackDepth<OrderedFloat<f64>>),
     Flush(Flush<OrderedFloat<f64>>),
+    Print(Print<OrderedFloat<f64>>),
+    PrintLn(PrintLn<OrderedFloat<f64>>),
 
+    // Arithmetic instructions)
```
JustusFluegel, line 36: Weird closing `)`

packages/push/src/instruction/printing/char.rs
```diff
@@ -0,0 +1,150 @@
+use std::io::Write;
+
+use super::super::{Instruction, instruction_error::PushInstructionError};
+use crate::{error::InstructionResult, instruction::NumOpens, push_vm::push_io::HasStdout};
+
+/// An instruction that "prints" a single character.
```
JustusFluegel, line 6: maybe `single, predetermined character` to avoid confusion with printing a character from the char stack?

packages/push/src/instruction/printing/char.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// This currently panics (due to an `unwrap()`) if attempting
+/// to write the character fails.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct PrintChar<const CHAR: char>;
+
+impl<const CHAR: char> PrintChar<CHAR> {
+    #[must_use]
+    pub const fn new() -> Self {
+        Self {}
```
JustusFluegel, line 54: I don't think the `{}` are required here.

packages/push/src/instruction/printing/char.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+impl<State, const CHAR: char> Instruction<State> for PrintChar<CHAR>
+where
+    State: HasStdout,
+{
+    type Error = PushInstructionError;
+
+    fn perform(&self, mut state: State) -> InstructionResult<State, Self::Error> {
+        // We need to remove this `unwrap()`.
```
JustusFluegel, line 65: Do we have an issue for this already? something like "Error handling for IO instructions?"

packages/push/src/instruction/printing/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+///    - `T` stack
+///      - One value
+///
+/// # Behavior
+///
+/// The `Print<T>` instruction "prints" the top value of
+/// the stack of type `T` to an internal buffer in the state.
+///
+/// This does _not_ put a newline after the printed value;
+/// see `PrintLn` for that behavior.
```
JustusFluegel, line 31: I would somehow note that this uses rust's `Display` implementation on the stack type to do the conversion to a string to print. We could reasonably have a PrintDebug as well, although I don't know if that would be that useful.

packages/push/src/instruction/printing/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+/// error when the `Stack<T>` is empty.
+///
+/// # Panics
+///
+/// This currently panics (due to an `unwrap()`) if attempting
+/// to write the character fails.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub struct Print<T> {
+    pub(crate) _p: PhantomData<T>,
+}
```
JustusFluegel, line 91: Add a new for this as well please.

packages/push/src/instruction/printing/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+///      - One value
+///
+/// # Behavior
+///
+/// The `PrintLn<T>` instruction "prints" the top value of
+/// the stack of type `T` to an internal buffer in the state,
+/// followed by a newline.
+///
+/// See `Print` for similar behavior that does not automatically
+/// include a newline after the value.
```
JustusFluegel, line 135: Same as above here, somehow note that this uses the `Display` trait.

packages/push/src/instruction/printing/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+///     .with_instruction_step_limit(10)
+///     .build();
+/// // Print the value on the integer stack.
+/// let mut result = PrintLn::<i64>::default().perform(push_state).unwrap();
+/// // Assert that the integer stack is now empty.
+/// assert_eq!(result.stack::<i64>().size(), 0);
+/// // Extract the printed output.
+/// let output = result.stdout_string().unwrap();
+/// // Assert that this is equal to the printed version of the one value
+/// // that was on the integer stack, i.e., `"42"`.
```
JustusFluegel, line 177: should the comment explain the newline?

packages/push/src/instruction/printing/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+/// error when the `Stack<T>` is empty.
+///
+/// # Panics
+///
+/// This currently panics (due to an `unwrap()`) if attempting
+/// to write the character fails.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub struct PrintLn<T> {
+    pub(crate) _p: PhantomData<T>,
+}
```
JustusFluegel, line 194: Another `new` would be great here!

packages/push/src/instruction/printing/string.rs
```diff
@@ -0,0 +1,125 @@
+use std::io::Write;
+
+use super::super::{Instruction, instruction_error::PushInstructionError};
+use crate::{error::InstructionResult, instruction::NumOpens, push_vm::push_io::HasStdout};
+
+/// An instruction that "prints" a specified string.
```
JustusFluegel, line 6: also the predetermined wording here, maybe?

packages/push/src/instruction/printing/string.rs
```diff
@@ -0,0 +1,125 @@
...
+/// let output = result.stdout_string().unwrap();
+/// // Assert that this is equal to "Hello".
+/// assert_eq!(output, "Hello");
+/// ```
+/// # Panics
+///
+/// This currently panics (due to an `unwrap()`) if attempting
+/// to write the character fails.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct PrintString(pub String);
```
JustusFluegel, line 57: maybe also a `new` here? I know that it isn't really needed since the field is pub and as such the constructor fn is also pub but might be nice for consistency with the other instructions.

packages/push/src/push_vm/push_io/mod.rs
```diff
@@ -0,0 +1,15 @@
+use std::{io::Write, string::FromUtf8Error};
+
+pub trait HasStdout {
+    type Stdout: Write;
+
+    fn stdout(&mut self) -> &mut Self::Stdout;
```
JustusFluegel, line 6: Maybe add some docs to this method?

packages/push/src/push_vm/push_io/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+
+    /// Return the contents of `stdout` as a `String`
+    ///
+    /// # Errors
+    ///
+    /// Returns a `FromUtf8Error` if there is a problem converting
+    /// the contents of `Self::Stdout` into a `String`.
+    fn stdout_string(&mut self) -> Result<String, FromUtf8Error>;
```
JustusFluegel, line 14: if this is in this trait I would provide a default impl here
```suggestion
    fn stdout_string(&mut self) -> std::io::Result<String>
    where
        Self::Stdout: Seek
    {
           self.stdout().rewind()?;
           let mut buf = String::new();
           self.stdout().read_to_string(&mut buf)?;
           
           Ok(buf)
    }
```

packages/push/src/push_vm/push_state.rs
```diff
@@ -128,6 +133,27 @@ impl State for PushState {
...
+impl HasStdout for PushState {
+    type Stdout = Cursor<Vec<u8>>;
+
+    fn stdout(&mut self) -> &mut Self::Stdout {
+        &mut self.stdout
+    }
+
+    // TODO: Can we avoid the `clone()` here? It's not a huge deal (we'll
+    //   only call it once at the end of running the program), but it would
+    //   still be nice to avoid it if possible.
```
JustusFluegel, line 145: No we can't, since the String later re-uses the cloned vec as it's internal buffer, which shouldn't depend on the current internal buffer (or can't)

JustusFluegel (changes requested):

packages/push/examples/number_io/main.rs
NicMcPhee: That's because it _is_ wrong. 😜

packages/push/examples/number_io/main.rs
NicMcPhee: Nope – good catch.

packages/push/src/instruction/printing/char.rs
NicMcPhee: We do now (#366).

packages/push/src/push_vm/push_io/mod.rs
NicMcPhee: I've added something here, but I'm not 100% convinced that this is terribly clear or useful. Thoughts?

packages/push/src/push_vm/push_io/mod.rs
NicMcPhee: Do we need something this complex that depends on `Seek`? Would it make more sense to just make the default implementation depend on `Cursor` and basically just be the one that we have in `PushState`?

packages/push/src/push_vm/push_state.rs
NicMcPhee: Gotcha, thanks for clarifying that.

packages/push/src/instruction/printing/mod.rs
NicMcPhee: I added it up in the "introduction" as a new 2nd paragraph.

NicMcPhee: We could have a `PrintDebug`, but I don't think we need it _now_. Maybe create an issue for it, but I'm not sure how important it is.

packages/push/src/instruction/printing/mod.rs
NicMcPhee: Yes, it probably should!

packages/push/src/push_vm/push_io/mod.rs
NicMcPhee: We ended up removing `stdout_string` from the trait, so this became a non-issue.

packages/push/examples/number_io/main.rs
```diff
@@ -0,0 +1,258 @@
...
+use strsim::damerau_levenshtein;
+use strum::IntoEnumIterator;
+
+use crate::args::{CliArgs, RunModel};
+
+// An input for this problem is a tuple of four `i64`s.
+#[derive(Debug, Copy, Clone)]
+struct Input {
+    i: i64,
+    f: f64,
```
JustusFluegel, line 43: Are there better names for these?

packages/push/examples/number_io/main.rs
```diff
@@ -0,0 +1,258 @@
...
+use rand::{
+    distr::{Distribution, Uniform},
+    rng,
+};
+use strsim::damerau_levenshtein;
+use strum::IntoEnumIterator;
+
+use crate::args::{CliArgs, RunModel};
+
+// An input for this problem is a tuple of four `i64`s.
```
JustusFluegel, line 39: And this is wrong.

packages/push/examples/number_io/main.rs
```diff
@@ -0,0 +1,258 @@
...
+
+impl Distribution<Input> for (Uniform<i64>, Uniform<f64>) {
+    fn sample<R: rand::prelude::Rng + ?Sized>(&self, rng: &mut R) -> Input {
+        let i = self.0.sample(rng);
+        let f = self.1.sample(rng);
+        Input { i, f }
+    }
+}
+
+// An output for this problem is an `i64`.
```
JustusFluegel, line 74: Same here

packages/push/examples/number_io/main.rs
```diff
@@ -0,0 +1,258 @@
...
+        penalty_value,
+        ..
+    } = CliArgs::parse();
+
+    let mut rng = rng();
+
+    #[expect(
+        clippy::as_conversions,
+        clippy::cast_precision_loss,
+        reason = "I'm OK with just converting these with `as` for now."
```
JustusFluegel, line 110: why? ", since i am fine with loosing the precision here"

packages/push/examples/number_io/main.rs
```diff
@@ -0,0 +1,258 @@
...
+        println!("Stdout for input {first_input}: {output}");
+
+        if best.test_results.total_result.0 == 0 {
+            println!("SUCCESS");
+            break;
+        }
+    }
+
+    println!(
+        "DL with no newline: {}",
```
JustusFluegel, line 187: is `DL` a common shorthand?

JustusFluegel (changes requested):

JustusFluegel (approved):

PR #367, opened 2025-04-13
Create creation functions for all instructions

JustusFluegel:
Makes it more straightforward to use them


NicMcPhee (approved):
This is great (and was probably kinda tedious). Thanks!

PR #373, opened 2025-04-29
Binary comparison operators don't consume enough arguments

### LINKED ISSUE #372: Binary comparison operators don't consume enough arguments for `IntInstruction` and `FloatInstruction`
All the binary comparison operators in `IntInstruction` and `FloatInstruction` (e.g., `IntInstruction::Equals` and `FloatInstruction::GreaterThan`) only consume _one_ input instead of two, which is defiitely a serious bug.

It'll be an easy fix (replacing `1` with `2` in a bunch of `.replace_on()` calls).

Going through an evolved "solution" to the Median problem turned up the problem. It used `IntInstruction::GreatThanEquals`:

```rust
Self::GreaterThanEqual => int_stack
   .top2()
   .map_err(PushInstructionError::from)
   .map(|(&x, &y)| x >= y)
   .push_onto(state)
   .with_stack_discard::<i64>(1),
```

The problem is the last like, where we're only discarding one value from the integer stack instead of 2.

Ultimately the _real_ solution will be replacing all those "hand rolled" implementations of comparison operators with the new generic `cmp` operators. These actually have tests and are (hopefully) correct. I checked at least one of them by hand in the example that brought all this up and it did fix the problem.

NicMcPhee:
This PR fixes a bug in the binary comparison operators for integers and floats by ensuring they consume two stack arguments instead of one. It also adds tests to capture the median calculation behavior in response to this correction.

- Fixes the argument consumption for binary comparison operators in IntInstruction and FloatInstruction.
- Adds new tests in tests/median.rs to capture and prevent regressions.
- Updates the integer and float modules to discard two values from the stack.

All the binary comparison operators in `IntInstruction` and `FloatInstruction` (e.g., `IntInstruction::Equals` and `FloatInstruction::GreaterThan`) only consume _one_ input instead of two, which is defiitely a serious bug.

It was an easy fix (replacing `1` with `2` in a bunch of `.replace_on()` calls).

The bugs with the integer and float comparison operators were turned up by trying to make sense of an evolved "solution" to the Median problem that didn't make sense.

I created the file `tests/median.rs` to help me debug the problem, and then to capture the issue in case I regress on this at some future point. I put this in the `tests/` folder since it wasn't clear where else it would go that would make more sense.

The "real" fix would be to convert both the int and float comparisons to use the new generic `cmp` instructions, but for now I'm just going to replace the 1s with 2s in the appropriate places.

Closes #372


JustusFluegel (approved):

NicMcPhee:
Thanks for the quick review @JustusFluegel!

PR #375, opened 2025-06-08
Add `Umad` constructor with balanced deletion rate

### LINKED ISSUE #286: Align our UMAD implementation with that in Clojush
Our current UMAD implementation isn't quite the same as that used in Clojush or as described in [the original UMAD paper](https://dl.acm.org/doi/10.1145/3205455.3205603). [In the Clojush implementation](https://github.com/lspector/Clojush/blob/e2c9d8c830715f7d1e644f6205c192b9e5ceead2/src/clojush/pushgp/genetic_operators.clj#L605) you can just specify addition rate and the deletion rate is computed from that in a way that ensures that the resulting sequence is on average the same length as the "parent" sequence.

We don't support that approach, and we should.

We currently require that users provide both the addition and deletion rates, and in our examples we typically use the same values for both. This actually leads to a slight bias towards shrinking programs, which makes our results not directly comparable to Clojush results.

## Why the bias?

Imagine that you have a sequence of 100 items (e.g., instructions) and an addition and deletion rate of 0.1. Then we would (on average) expect to add 10 items in the addition phase, leading to a total of 110 items. If we then delete with a rate of 0.1, then we would expect to delete (on average) 11 items, leading to a sequence of just 99 items.

## How do we fix it?

If we set the deletion rate $D$ to be $A / (1 + A)$ then on average the additions and deletions should cancel out. This is because if we start with a sequence of length $l$ we expect the length after first adding and then deleting to be:

$$
(l \times (1 + A)) \times (1 - D)
$$ 

For this to be $l$ again, we need $(1 + A) \times (1 - D) = 1$:

$$
\begin{eqnarray}
(1 + A) \times (1 - D) & = & (1 + A) \times (1 - A / (1 + A)) \\
& = & 1 + A - (1 + A) \times A / (1 + A) \\
& = & 1 + A - A \\
& = & 1
\end{eqnarray}
$$

## But maybe not?

All that said, our implementation of UMAD is quite different than that used in Clojush thanks to some nice "tricks" suggested by @MizardX. We probably want to make sure that our implementation is actually using the addition and deletion rates in the same way as those in Clojush, and that using $A / (1 + A)$ for the deletion rate does in fact lead to the addition and deletion cancelling each other out.

NicMcPhee:
This adds a new constructor for `Umad` mutation operator which calculates the balanced deletion based on the given addition rate.

UMAD, as described in ["Program Synthesis using Uniform Mutation by Addition and Deletion"](https://doi.org/10.1145/3205455.3205603), recommends using a balanced deletion rate which is

$$
\textrm{deletion rate} = \textrm{addition rate} / (1 + \textrm{addition rate})
$$

If you use this deletion rate, this ensures that, on average, genomes maintain the same length after mutation.

This PR also updates the Push examples to use the new `Umad::new_with_balanced_deletion()` constructor.

Closes #286

### DIFF
--- a/packages/ec-linear/examples/hiff/main.rs
+++ b/packages/ec-linear/examples/hiff/main.rs
@@ -52,9 +52,7 @@ fn hiff(bits: &[bool]) -> (bool, TestResults<Score<usize>>) {
                 .results
                 .into_iter()
                 .chain(right_score.results)
-                .chain(once(Score::from(
-                    all_same.then_some(len).unwrap_or_default(),
-                )))
+                .chain(once(Score::from(if all_same { len } else { 0 })))
                 .collect(),
         )
     }


JustusFluegel (dismissed):
LGTM! :)

JustusFluegel (approved):
LGTM Take 2

PR #379, opened 2025-06-15
Further align our `Umad` implementation with the UMAD paper

NicMcPhee:
[The original UMAD paper](https://doi.org/10.1145/3205455.3205603) randomly decides whether to put a newly added gene on the left or right side of the current gene. We didn't actually provide that functionality, and this fixes that.

Hopefully our UMAD implementation now matches (or can match with the correct parameters) the published definition.

While we were here, we also added documentation and examples for `Umad`.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,6 +10,7 @@
         "doctest",
         "doctests",
         "esitsu",
+        "Freitag",
         "GECCO",
         "genericize",
         "Goues",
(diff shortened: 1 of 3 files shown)


NicMcPhee:
@JustusFluegel When you get a chance to review this, feel free to Clippy doctests issues again. If we don't deal with those during the week, we can make them the point of a stream next weekend.

JustusFluegel (approved):

PR #380, opened 2025-06-16
Add bacon config to simplify doc writing and updating

JustusFluegel:
This adds a `doc` and `private-doc` job to bacon which will automatically run cargo doc and open the docs, then switch to a `*-watch` task and update the docs as they are changed.

The keybinds `d` (and `alt-d` for private)  are also provided within bacon.

The difference between the `private-*` version and non-private version of the doc job is whether  they include the `--document-private-items` flag, the private job does, the non-private one doesn't.

### DIFF
--- a/bacon.toml
+++ b/bacon.toml
@@ -21,6 +21,8 @@ command = [
 ]
 on_success = "job:doc-watch"
 need_stdout = false
+allow_failures = true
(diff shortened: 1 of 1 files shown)


NicMcPhee (approved):
This is very nice – thanks!

PR #387, opened 2025-07-21
Improve lexicase sampling

JustusFluegel:
This lazily selects a random test case to go by instead of shuffling up front. This reduces rng calls a bunch when not all test cases are needed to select a "winner".

In case multiple winners are found, this also randomly chooses one instead of shuffling them and choosing the first one.


NicMcPhee (approved):
This looks wonderful – thanks a ton for speeding things up!

PR #392, opened 2025-08-10
Remove `Vector<T>` type

### LINKED ISSUE #364: Remove `Vector`
We should figure out if we have any need for `Vector`, and if not get rid of it. We might just want to implement `Linear` for `Vec<T>` and not have the `Vector<T>` wrapper.

Another option would be to keep `Vector<T>` but rename it as `LinearGenome<T>`. The `Vector` might be read as implying fixed length genomes, when the actual implementation is (and should be) more general than that.

NicMcPhee:
The `Vector<T>` type wasn't very helpful and didn't really do anything that `Vec<T>` couldn't do. This removes that, and implemented `Genome` and `Linear` for `Vec<T>`, replacing the small number of uses of `Vector<T>` with `Vec<T>`.

Closes #364


JustusFluegel (approved):
LGTM

PR #405, opened 2025-11-02
Fix ci breakages

JustusFluegel:
This includes:

- a false positive rust lint: https://github.com/rust-lang/rust/issues/147648
- rustdocs --nocapture rename: https://github.com/rust-lang/rust/pull/148115
- actually erroring on warnings with clippy for doctests


NicMcPhee (approved):
Thanks for grinding through all that!

PR #416, opened 2025-12-07
Remove child maker

### LINKED ISSUE #413: Remove `ChildMaker`
The trait isn't actually used anymore and it's just there as legacy, so we should remove it.

JustusFluegel:
Somewhere _early_ in the project Nic had envisioned `ChildMaker` as a thing that would encapsulate the process of making new children given a population and a selector. In the end, that never got used because we instead just built up an `Operator` that did the same thing and never needed the trait. The only place where this would have been used is in `Generation` (a problematic type in its own right), which is generic in `C` where `C` is essentially the "child maker". It is specified to be an `Operator`, however, that takes a population and returns an individual.

So we're removing the `ChildMaker` type. If we want something like this in the future, what would probably make more sense is a _builder_ that would take the various "common" components of a pipeline (the selector, the mutator, etc.) and build an operator. That would help hide the potential complexity of the whole `Operator` business, might provide better error messages, and not introduce a type that is never used.

This is stacked on top of #349 so that should be merged first, then this pr should be way more reasonable (see only the last commit for now)

Closes #413


NicMcPhee (dismissed):
This looks good – thanks for catching this.

I've added some notes to the PR description that you might want to look at before we actually merge this in.

PR #425, opened 2026-01-01
Remove lint ignores that are obsolete because they are fixed / implemented

JustusFluegel:
I fixed https://github.com/rust-lang/rust-clippy/issues/15168 in clippy and the notes about the lint reasons rfc landing in stable sometime are no longer relevant cuz they already are in stable for a while.


NicMcPhee (approved):
Thanks!

PR #428, opened 2026-01-04
Restructure performance modules

JustusFluegel:
tl;dr: Pulls out `Score`, `Error`, `TestResult` and `TestResults` into their own module under the `performance` super module.

---

> [!note] 
> [Created by Google Gemini](https://github.com/unhindered-ec/unhindered-ec/pull/428#issuecomment-3765457374)

### Highlights

* **Module Restructuring**: The `Score`, `Error`, `TestResult`, and `TestResults` types have been moved from `ec-core/src/test_results.rs` into a new, dedicated `performance` super module (`ec-core/src/performance/`).
* **Type Renaming**: The `Score` and `Error` types have been renamed to `ScoreValue` and `ErrorValue` respectively, and are now defined in their own sub-modules (`score_value.rs` and `error_value.rs`) within the `performance` module.

### DIFF
--- a/packages/ec-linear/examples/count_ones/main.rs
+++ b/packages/ec-linear/examples/count_ones/main.rs
@@ -112,15 +112,14 @@ fn main() -> miette::Result<()> {
 
 #[cfg(test)]
 mod test {
-    use ec_core::performance::{self, TestResults};
+    use ec_core::performance::{score_value::ScoreValue, test_results::TestResults};
 
     use super::count_ones;
 
     #[test]
     fn non_empty() {
         let input = [false, true, true, true, false, true];
-        let output: TestResults<performance::ScoreValue<i64>> =
-            [0, 1, 1, 1, 0, 1].into_iter().collect();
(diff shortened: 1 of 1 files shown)


NicMcPhee:
I don't think the first two comments make sense for us, but the suggestions about cleaning up the imports of `ErrorValue` seem reasonable so I've done them.

packages/ec-linear/examples/count_ones/main.rs
```diff
@@ -109,18 +109,3 @@ fn main() -> miette::Result<()> {
...
-    use super::count_ones;
-
-    #[test]
-    fn non_empty() {
-        let input = [false, true, true, true, false, true];
-        let output: TestResults<test_results::Score<i64>> =
-            [0, 1, 1, 1, 0, 1].into_iter().collect();
-        assert_eq!(output, count_ones(&input));
-    }
-}
```
JustusFluegel: Why was this deleted?  I think this was testing the `count_ones` function.

NicMcPhee (approved):
Looks good!

PR #431, opened 2026-01-18
Part two of the patchset of #349

JustusFluegel:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,3 +18,6 @@ The following steps are usually required:
     2. Bump the version of all the workspace packages to the same version you have choosen in step 1. If they don't have a `version = "..."` key already, feel free to add them in **addition** to the `path = "..."` key. That will ensure that we still use path dependencies while developing locally and only use the version when publishing (cargo supports that).
     3. If you are publishing to a registry other than your system's default one (which is usually `crates.io`) you'll also need to add a `registry = "..."` key along the version key from the step above. Make sure to not push this (commit it only locally to satisfy cargo), as the  CI won't find your locally configured registry and as such will fail to build.
 3. Now we can finally use `cargo publish --workspace` to publish the workspace packages. Make sure that you include the `--registry` argument if you plan to publish to another registry than your systems default one.
+
+## Publishing v1.0.0
+Once we publish any of this to crates.io we need to replace some links with actual urls to docs.rs which aren't valid yet since the packages don't exist yet. For that search for  `(#)` in the code.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -407,6 +407,7 @@ name = "ec-core"
 version = "0.1.0"
 dependencies = [
  "criterion",
+ "ec-linear",
  "ec_macros",
  "itertools 0.14.0",
  "macro_railroad_annotation",
--- a/packages/ec-core/Cargo.toml
+++ b/packages/ec-core/Cargo.toml
@@ -38,6 +38,7 @@ thiserror = { workspace = true }
 
 [dev-dependencies]
 criterion = { workspace = true }
+ec-linear = { path = "../ec-linear" }
 miette = { workspace = true, features = ["fancy"] }
 proptest = { workspace = true }
 rand = { workspace = true, features = ["alloc", "small_rng"] }
--- a/packages/ec-core/src/generation.rs
+++ b/packages/ec-core/src/generation.rs
@@ -3,41 +3,524 @@ use rayon::prelude::{FromParallelIterator, ParallelIterator};
 
 use crate::{operator::Operator, population::Population};
 
-pub struct Generation<P, C> {
-    population: P,
+/// Collection of data about each iteration in the evolution process.
+///
+/// Notably, this includes the Population of the current Generation, as well as
+/// a child maker [`Operator`] which describes how to generate the next
+/// generation from this.
+///
+/// Take a look at [`Generation::par_next`] and [`Generation::serial_next`] for
+/// further information on how this is used to generate new generations.
+///
+/// # Example[^ec-linear-usage]
+/// ```
+/// # use ec_core::{
+/// #     performance::{score_value::ScoreValue, test_results::TestResults},
+/// #     individual::{scorer::FnScorer, ec::WithScorer},
+/// #     operator::{
+/// #         selector::{lexicase::Lexicase, Select},
+/// #         genome_extractor::GenomeExtractor,
+/// #         recombinator::Recombine,
+/// #         mutator::Mutate,
+/// #         genome_scorer::GenomeScorer,
+/// #         Composable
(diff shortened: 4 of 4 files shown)


packages/ec-core/src/operator/composable/mod.rs
```diff
@@ -1,14 +1,14 @@
...
 mod map;
 mod repeat_with;
 mod then;
 
+pub use and::*;
 // derive macro
 pub use ec_macros::Composable;
+pub use map::*;
+pub use repeat_with::*;
+pub use then::*;
```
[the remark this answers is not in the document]
JustusFluegel: nope rustfmt doesn't like it

packages/ec-core/src/performance/error_value.rs
```diff
@@ -12,12 +31,45 @@ pub struct ErrorValue<T>(pub T);
 static_assertions::assert_impl_all!(ErrorValue<()>: Clone);
 
 impl<T: Ord> Ord for ErrorValue<T> {
+    /// Compares two errors.
+    ///
+    /// Errors are ordered in reverse wrt.
```
NicMcPhee:
```suggestion
    /// Errors are ordered in reverse wrt
```

packages/ec-core/src/performance/error_value.rs
```diff
@@ -26,12 +78,32 @@ impl<T: PartialOrd> PartialOrd for ErrorValue<T> {
 }
 
 impl<T: PartialOrd> PartialOrd<T> for ErrorValue<T> {
+    /// Compares the value of an error
+    ///
+    /// In contrast to comparing two errors, this is *not* ordered in reverse.
```
NicMcPhee: Should we say something about *why* this is not the same as when we're comparing errors? My memory is that there was a reasonable argument for this, but I don't remember what it was off the top of my head. It might be useful to document that here.

packages/ec-core/src/performance/error_value.rs
```diff
@@ -45,19 +117,50 @@ impl<T: Display> Display for ErrorValue<T> {
...
+    /// #
+    /// assert_eq!(ErrorValue::from(5), ErrorValue(5));
+    /// ```
     fn from(error: T) -> Self {
         Self(error)
     }
 }
 
 impl<T: Sum> Sum<T> for ErrorValue<T> {
+    /// Create a new [`ErrorValue`] from summing up a iterator of values.
```
NicMcPhee:
```suggestion
    /// Create a new [`ErrorValue`] from summing up an iterator of values.
```

packages/ec-core/src/performance/score_value.rs
```diff
@@ -66,6 +148,24 @@ where
     T: ToOwned,
     Self: Sum<<T as ToOwned>::Owned>,
 {
+    /// Sum up a Iterator of references to [`ScoreValue`]'s.
```
NicMcPhee: Same question about saying something about `T` typically being clonable here?

packages/ec-core/src/performance/test_result.rs
```diff
@@ -2,43 +2,162 @@ use std::cmp::Ordering;
...
+/// if comparing two [`TestResult::Error`] variants, smaller is better.
+///
+/// If you don't need the ability to have both bigger-is-better and
+/// smaller-is-better semantics, take a look at using
+///
+/// - [`ScoreValue`] (bigger is better), or
+/// - [`ErrorValue`] (smaller is better)
+///
+/// directly instead, which are probably better suited for your needs (and
+/// as a bonus also implement [`Ord`].)
```
NicMcPhee: Nice – thanks.

packages/ec-core/src/performance/test_result.rs
```diff
@@ -2,43 +2,162 @@ use std::cmp::Ordering;
...
+///
+/// # Examples
+/// ```
+/// # use ec_core::performance::test_result::TestResult;
+/// # use std::cmp::Ordering;
+/// assert_eq!(
+///     TestResult::<i32, i32>::score(5).partial_cmp(&TestResult::<i32, i32>::score(10)),
```
NicMcPhee: I think this line (and similar lines below) would be a lot easier to read if we gave names to the two components (like maybe `score_5` and `score_10`) and then compared them.

packages/ec-core/src/performance/test_results.rs
```diff
@@ -1,5 +1,18 @@
...
+/// result.
+///
+/// # Example
+/// ```
+/// # use ec_core::performance::{test_results::TestResults, score_value::ScoreValue};
+/// let results = [ScoreValue(5), ScoreValue(10)]
+///     .into_iter()
+///     .collect::<TestResults<_>>();
+///
+/// assert_eq!(results, ScoreValue(15));
```
NicMcPhee: This is an odd assertion to use as the example here, because it makes a `TestResults` value look "just" like a `ScoreValue`, obscuring the fact that it still has all the individual results. Can we expand this so that the structure of a `TestResults` is clearer?

NicMcPhee (changes requested):
Thanks again for all this work! Most of my comments/questions are minor, and hopefully we can deal with them in tomorrow's stream.

packages/ec-core/src/genome.rs
```diff
@@ -1,8 +1,34 @@
...
+///
+/// # Example
+/// This is how you might implement a Bitstring genome type:
+/// ```
+/// # use ec_core::genome::Genome;
+/// # #[allow(dead_code)]
+/// struct Bitstring {
+///     length: usize,
+///     storage: Vec<u64>,
+/// }
```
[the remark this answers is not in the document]
JustusFluegel: I wanted something involved than just a simple `Vec<bool>` as an example. But how about I rename this example to `PackedBitstring` to be more explicit in contrasting the existing `Bitstring`?

packages/ec-core/src/performance/test_result.rs
```diff
@@ -2,43 +2,162 @@ use std::cmp::Ordering;
...
+    /// ```
     #[must_use]
     pub const fn error(error: E) -> Self {
         Self::Error(ErrorValue(error))
     }
 }
 
 impl<S: Default, E> Default for TestResult<S, E> {
+    /// Create a new [`TestResult`], defaulting to
+    /// [`TestResult::Score(ScoreValue::default())`](TestResult::Score)
```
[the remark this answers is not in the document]
JustusFluegel: It can be useful to have a default implementation if you just need "some value" of the given type; as such it is usually implemented to eagerly implement the `Default` trait.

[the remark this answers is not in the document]
NicMcPhee: Maybe something like:

```rust
/// This decision is somewhat arbitrary, but ensures that users of 
///`TestResult` have an implementation of `Default` if they need one.
```

Issue #429, opened 2026-01-18
Consolidate `ChooseCloing` and `ChooseCloningOwned`


JustusFluegel:
Historically `ChooseCloning` existed first since it is a thin wrapper around `rand::distr::Choose` and we later added `ChooseCloningOwned` but now that both exist we might want to consider consolidating it into a single distribution helper `ChooseCloning` since (I think) the current `ChooseCloning` could just be a `type ChooseCloning<'a, T> = ChooseCloningOwned<&'a [T], T>`.  

We should do some research on that and see if that'd actually work.

NicMcPhee:
We could also consider changing the name `ChooseCloning` to `ChooseWithCloning` if we were to make this change.

PR #432, opened 2026-01-18
Part three of the patchset of #349

JustusFluegel:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,4 +20,4 @@ The following steps are usually required:
 3. Now we can finally use `cargo publish --workspace` to publish the workspace packages. Make sure that you include the `--registry` argument if you plan to publish to another registry than your systems default one.
 
 ## Publishing v1.0.0
-Once we publish any of this to crates.io we need to replace some links with actual urls to docs.rs which aren't valid yet since the packages don't exist yet. For that search for  `(#)` in the code.
+Once we publish any of this to crates.io we need to replace some placeholder links with actual URLs to docs.rs. These placeholders are not yet valid because the packages don't exist in the registry. To find them, search for  `(#)` in the code.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,4 +20,4 @@ The following steps are usually required:
 3. Now we can finally use `cargo publish --workspace` to publish the workspace packages. Make sure that you include the `--registry` argument if you plan to publish to another registry than your systems default one.
 
 ## Publishing v1.0.0
-Once we publish any of this to crates.io we need to replace some placeholder links with actual URLs to docs.rs. These placeholders are not yet valid because the packages don't exist in the registry. To find them, search for  `(#)` in the code.
+Once we publish any of this to crates.io we need to replace some placeholder links with actual URLs to docs.rs. These placeholders are not yet valid because the packages don't exist in the registry. To find them, search for `]: #` in the code.
--- a/packages/ec-core/src/distributions/conversion.rs
+++ b/packages/ec-core/src/distributions/conversion.rs
@@ -5,29 +5,122 @@ use rand::{
 
 use super::wrappers::{choose_cloning::ChooseCloning, choose_cloning_owning::ChooseCloningOwning};
 
+/// Conversion into a [`Distribution`].
+///
+/// # Notable implementations
+/// - `[T; N]`
+/// - `&'a [T]`
+/// - `&'a [T; N]`
+///
+/// These implementations allow you to easily convert a list of elements into a
+/// [`Distribution`] you can sample one from. Usually, these implementations are
+/// meant to turn types into Uniform distributions.
+///
+/// # Example
+/// ```
+/// # use ec_core::distributions::conversion::IntoDistribution;
+/// # use rand::{rng, distr::{Distribution, slice::Empty}};
+/// #
+/// # fn main() -> Result<(),Empty> {
+/// let my_collection = [1i32, 2, 3, 4, 5];
+/// let my_distribution = my_collection.into_distribution()?;
+///
+/// let my_sample = my_distribution.sample(&mut rng());
+/// assert!(my_collection.contains(&my_sample));
+/// # let _ = my_sample;
+/// # Ok(())
+/// # }
+/// ```
 pub trait IntoDistribution<Element> {
     type Distribution: Distribution<Element>;
     type Error;
 
-    /// Converts some type into a Distribution (fallible)
+    /// Converts self into a [`Distribution`] by moving.
+    ///
+    /// This is usually implemented for collections and then allows uniformly
+    /// sampling from them.
+    ///
+    /// # Example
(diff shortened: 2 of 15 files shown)


CONTRIBUTING.md
```diff
@@ -18,3 +18,6 @@ The following steps are usually required:
...
     3. If you are publishing to a registry other than your system's default one (which is usually `crates.io`) you'll also need to add a `registry = "..."` key along the version key from the step above. Make sure to not push this (commit it only locally to satisfy cargo), as the  CI won't find your locally configured registry and as such will fail to build.
 3. Now we can finally use `cargo publish --workspace` to publish the workspace packages. Make sure that you include the `--registry` argument if you plan to publish to another registry than your systems default one.
+
+## Publishing v1.0.0
+Once we publish any of this to crates.io we need to replace some placeholder links with actual URLs to docs.rs. These placeholders are not yet vailid because the packages don't exist in the registry. To find them, search for `]: #` in the code.
```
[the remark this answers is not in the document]
JustusFluegel: this is already mentioned in part 2.

packages/ec-core/src/distributions/conversion.rs
```diff
@@ -46,71 +139,179 @@ where
...
 }
 
-impl<'a, U> IntoDistribution<U> for &'a Vec<U>
+impl<'a, T> IntoDistribution<T> for &'a Vec<T>
 where
-    U: Clone,
+    T: Clone,
 {
-    type Distribution = ChooseCloning<'a, U>;
+    type Distribution = ChooseCloning<'a, T>;
```
NicMcPhee: I know it's not something for *this* PR, but I wonder if something like `ChooseWithCloning` would be a better name for things like this. As it is it sounds like we're choosing to use cloning as opposed to some other strategy.

NicMcPhee (commented):
I'm down to just a few files, but I figured I'd submit this as a comment to make my (very few) notes visible.

packages/ec-core/src/distributions/conversion.rs
JustusFluegel: #429 might also be relevant

packages/ec-core/src/operator/mod.rs
```diff
@@ -23,7 +23,45 @@ pub mod selector;
...
+///
+/// For more information on how data pipelines work in general in this crate,
+/// see the [module-level documentation](super::operator)
+///
+/// For concrete operators that you might wish to use can take a look at
+/// [`Map`](composable::Map) and [`Then`](composable::Then) among
+/// others.
+///
+/// # Example
+/// This example shows how to implement a [`Operator`], not how you use
```
NicMcPhee:
```suggestion
/// This example shows how to implement a [`Operator`] (in this case `Apply`, which applies an encapsulated unary function), not how you use
```

packages/ec-core/src/operator/mod.rs
```diff
@@ -35,9 +73,28 @@ pub use composable::Composable;
...
+    ///
+    /// This also takes an rng that is passed along to customize random number
+    /// generation behavior and avoid re-creating RNGs in each operator.
+    ///
+    /// # Example
+    /// ```
+    /// # use ec_core::operator::{constant::Constant, Operator};
+    /// # use rand::rng;
+    /// #
+    /// let my_constant_operator: Constant<_> = Constant::new(5);
```
NicMcPhee, line 95:
```suggestion
    /// // A _constant_ operator that will always return the same value
    /// // (in this case 5) regardless of its input.
    /// let my_constant_operator: Constant<_> = Constant::new(5);
```

packages/ec-core/src/population.rs
```diff
@@ -1,10 +1,52 @@
...
+///
+/// This trait is auto-implemented for all collections with both the `&T:
+/// IntoIterator<Item = &Individual>` trait and the `T: IntoIterator<Item =
+/// Individual>` trait implemented.
+///
+/// # Example
+/// ```
+/// # use ec_core::{individual::ec::EcIndividual, population::Population};
+/// #
+/// let my_population = vec![
```
NicMcPhee, line 12: Should we make it clear that `my_population` is being a `Population` here, and not a `Vec`? I realize that's implied in the call to `check_not_empty()` below, but I'd think that in documentation we'd want to be as explicit as possible?

```suggestion
/// let my_population: Population = vec![
```

packages/ec-core/src/population.rs
```diff
@@ -1,10 +1,52 @@
...
+    /// The type of the individual of this population
     type Individual;
 
+    /// Returns `true` if this Population doesn't contain any individuals
+    ///
+    /// # Example
+    /// ```
+    /// # use ec_core::population::Population;
+    /// #
+    /// let my_population = vec![1, 2, 3];
```
NicMcPhee, line 35: Maybe the previous comment applies here (and below) as well? I'm somewhat less concerned about it down here, but I think it might still be a good idea.

packages/ec-linear/src/mutator/with_rate.rs
```diff
@@ -5,11 +5,47 @@ use rand::Rng;
...
+/// #     mutator::with_rate::WithRate,
+/// #     genome::bitstring::Bitstring
+/// # };
+/// # use ec_core::operator::mutator::Mutator;
+/// #
+/// let mut rng = rand::rng();
+///
+/// let genome = Bitstring::random(100, &mut rng);
+///
+/// let mutator = WithRate::new(1.0);
```
NicMcPhee, line 24:
```suggestion
/// // Flip bits with probability 1.0, i.e., flip every bit 
/// let mutator = WithRate::new(1.0);
```

packages/ec-linear/src/mutator/with_rate.rs
```diff
@@ -5,11 +5,47 @@ use rand::Rng;
...
 }
 
+impl WithRate {
+    /// Create a new [`WithRate`] mutator from the given mutation rate
+    /// (probability)
+    ///
+    /// # Example
+    /// ```
+    /// # use ec_linear::mutator::with_rate::WithRate;
+    /// let mutator = WithRate::new(0.5);
```
NicMcPhee, line 43:
```suggestion
    /// // Flip each bit with a probability of 0.5,
    /// // i.e., flip on average half the bits
    /// let mutator = WithRate::new(0.5);
```

NicMcPhee (changes requested):

NicMcPhee:
This is great. I made a few suggestions, but they're all minor. I really appreciate this work!

packages/ec-core/src/population.rs
JustusFluegel: well Population is a trait and not a value, so it would need to be 
```suggestion
/// let my_population: impl Population = vec![
```
which is sadly not supported yet (see https://github.com/rust-lang/rust/issues/63065 )

I added a comment that mentions that `Vec<T>` implements Population.

NicMcPhee (approved):
Thanks for all this excellent work!

PR #433, opened 2026-01-18
Part four of the patchset of #349

JustusFluegel:


packages/ec-core/src/distributions/wrappers/choose_cloning.rs
```diff
@@ -53,6 +87,28 @@ impl<T> Distribution<T> for ChooseCloning<'_, T>
...
+    /// # Example
+    /// ```
+    /// # use rand::{rng, distr::{Distribution, slice::Empty}};
+    /// # use ec_core::distributions::wrappers::choose_cloning::ChooseCloning;
+    /// #
+    /// # fn main() -> Result<(), Empty> {
+    /// let collection = [vec![0], vec![1], vec![2], vec![3]];
+    /// let distribution = ChooseCloning::new(&collection)?;
+    ///
+    /// let sample = distribution.sample(&mut rng());
```
NicMcPhee:
```suggestion
    /// let sample: Vec<i32> = distribution.sample(&mut rng());
```

Since the assertion doesn't actually require that `sample` be an owned type rather than a reference, it seems good to include the type signature here to make the owning explicit.

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
```diff
@@ -14,6 +14,32 @@ use crate::distributions::finite::Finite;
 ///
 /// Also see [`ChooseCloning`](super::choose_cloning::ChooseCloning) for an
 /// alternative that borrows the collection instead.
+///
+/// # Generics
+///
+/// This distribution needs to validate that it is not empty in the new (or else
+/// it would not be a valid [`Distribution`]), as such the element `U` of the
+/// collection `T` needs to be determined at construction time and is part of
+/// this structs interface.
```
NicMcPhee:
```suggestion
/// this struct's interface.
```

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
```diff
@@ -14,6 +14,32 @@ use crate::distributions::finite::Finite;
 ///
 /// Also see [`ChooseCloning`](super::choose_cloning::ChooseCloning) for an
 /// alternative that borrows the collection instead.
+///
+/// # Generics
+///
+/// This distribution needs to validate that it is not empty in the new (or else
```
NicMcPhee:
```suggestion
/// This distribution needs to validate that it is not empty in the constructor `new` (or else
```

packages/ec-core/src/distributions/wrappers/choose_cloning.rs
NicMcPhee, line 38: Should the documentation say something about the role of the lifetime `'a` here? Since we're cloning, it might seem odd that we even need a lifetime here. We need it because the `Choose` type needs it, but I don't think that's totally obvious here.

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
```diff
@@ -14,6 +14,32 @@ use crate::distributions::finite::Finite;
...
+/// it would not be a valid [`Distribution`]), as such the element `U` of the
+/// collection `T` needs to be determined at construction time and is part of
```
NicMcPhee:
```suggestion
/// collection `T` needs to be known at construction time and is part of
```

Does this change make sense? Would "specified" be better?

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
```diff
@@ -14,6 +14,32 @@ use crate::distributions::finite::Finite;
...
+///
+/// # Example
+/// ```
+/// # use rand::{rng, distr::{Distribution, slice::Empty}};
+/// # use ec_core::distributions::wrappers::choose_cloning_owning::ChooseCloningOwning;
+/// #
+/// # fn main() -> Result<(), Empty> {
+/// let collection = [vec![0], vec![1], vec![2], vec![3]];
+/// //               collection gets moved here    \/
+/// let distribution = ChooseCloningOwning::new(collection)?;
```
NicMcPhee:
```suggestion
/// let distribution: ChooseCloningOwning<[Vec<i32>; 4], Vec<i32>> = ChooseCloningOwning::new(collection)?;
```

I wonder if an array of `String`s would be easier to read than array of `Vec`s? It might make the difference between the array and the values in the array clearer? I don't want to derail the PR over this, but I thought I'd mention it.

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
```diff
@@ -14,6 +14,32 @@ use crate::distributions::finite::Finite;
...
+/// // as such this is not valid
+/// // let _ = collection
+///
+/// let sample = distribution.sample(&mut rng());
```
NicMcPhee:
```suggestion
/// let sample: Vec<i32> = distribution.sample(&mut rng());
```

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
```diff
@@ -45,9 +71,9 @@ where
     /// # use rand::distr::slice::Empty;
     /// # use ec_core::distributions::wrappers::choose_cloning_owning::ChooseCloningOwning;
     /// #
-    /// let options = [1, 2, 3];
-    /// let distr = ChooseCloningOwning::new(options)?;
-    /// # let _ = distr;
+    /// let collection = [vec![0], vec![1], vec![2], vec![3]];
+    /// let distribution = ChooseCloningOwning::new(collection)?;
```
NicMcPhee:
```suggestion
    /// let distribution: ChooseCloningOwning<[Vec<i32>; 4], Vec<i32>> = ChooseCloningOwning::new(collection)?;
```

packages/ec-core/src/distributions/wrappers/choose_cloning.rs
```diff
@@ -36,6 +41,19 @@ impl<'a, T> ChooseCloning<'a, T> {
...
     ///
+    /// # Example
+    /// ```
+    /// # use rand::distr::slice::Empty;
+    /// # use ec_core::distributions::wrappers::choose_cloning::ChooseCloning;
+    /// #
+    /// # fn main() -> Result<(), Empty> {
+    /// let collection = [vec![0], vec![1], vec![2], vec![3]];
+    /// let distribution = ChooseCloning::new(&collection)?;
+    /// # let _ = distribution;
```
NicMcPhee, line 52: Would it make sense to add an underscore to the name of the variable and skip the `# let _ = …` lines?

```suggestion
    /// let _distribution = ChooseCloning::new(&collection)?;
```

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
```diff
@@ -68,6 +94,22 @@ where
...
+    /// [`Distribution`].
+    ///
+    /// # Example
+    /// ```
+    /// # use rand::distr::slice::Empty;
+    /// # use ec_core::distributions::{wrappers::choose_cloning_owning::ChooseCloningOwning, finite::Finite};
+    /// #
+    /// # fn main() -> Result<(), Empty> {
+    /// let collection = [vec![0], vec![1], vec![2], vec![3]];
+    /// let distribution = ChooseCloningOwning::new(collection)?;
```
NicMcPhee:
```suggestion
    /// let distribution: ChooseCloningOwning<[Vec<i32>; 4], Vec<i32>> = ChooseCloningOwning::new(collection)?;
```

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
```diff
@@ -78,6 +120,28 @@ where
...
+    /// 3. return a clone of that selected element
+    ///
+    /// # Example
+    /// ```
+    /// # use rand::{rng, distr::{Distribution, slice::Empty}};
+    /// # use ec_core::distributions::wrappers::choose_cloning_owning::ChooseCloningOwning;
+    /// #
+    /// # fn main() -> Result<(), Empty> {
+    /// let collection = [vec![0], vec![1], vec![2], vec![3]];
+    /// let distribution = ChooseCloningOwning::new(collection)?;
```
NicMcPhee:
```suggestion
    /// let distribution: ChooseCloningOwning<[Vec<i32>], Vec<i32>> = ChooseCloningOwning::new(collection)?;
```

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
```diff
@@ -78,6 +120,28 @@ where
...
+    ///
+    /// let sample = distribution.sample(&mut rng());
```
NicMcPhee:
```suggestion
    /// let sample: Vec<i32> = distribution.sample(&mut rng());
```

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
NicMcPhee:
```suggestion
    // We store the NonZeroUsize in the struct here, since we need to check this invariant in the
```

NicMcPhee, line 73: I feel like we've talked about this, but why are we using `Borrow` instead of `AsRef`? Is this a decision that should be documented here?

NicMcPhee:
```suggestion
            "This should never happen since the `new` method provides these guarantees"
```

NicMcPhee:
```suggestion
        // FIXME: Check the performance of this, and if necessary replace with
```

NicMcPhee:
```suggestion
            slice.len() > idx,
```

Won't `idx = len` lead to an out-of-bounds error when we call `get()` below? I realize that all the tests and checks should mean none of this should be an issue, but it seems that we want to get this assertion right regardless.

packages/ec-core/src/individual/ec.rs
```diff
@@ -16,10 +16,20 @@ use super::{
...
-/// Usually this is generated by an [`IndividualDistribution`] wrapping a
-/// genome generator.
+/// This might be generated by an [`IndividualDistribution`] wrapping a
+/// genome generator (for the initial population).
+///
+/// The test results `T` are usually either a simple value (`i32`, etc.), a
+/// [`ScoreValue`](ec_core::performance::score_value::ScoreValue) or
+/// [`ErrorValue`](ec_core::performance::error_value::ErrorValue) or a
+/// collection using a type like
+/// [`TestResults`](ec_core::performance::test_results::TestResults).
```
NicMcPhee, line 26:
```suggestion
/// The test results `T` are usually either 
///   - a simple value (`i32`, etc.),
///   - a [`ScoreValue`](ec_core::performance::score_value::ScoreValue) or [`ErrorValue`](ec_core::performance::error_value::ErrorValue), or 
///    - a collection using a type like [`TestResults`](ec_core::performance::test_results::TestResults).
```

We'll presumably want to change this to refer to `Accumulated` instead of `TestResults`, but that's probably part of the "Get rid of `TestResults`" PR.

packages/ec-core/src/individual/ec.rs
```diff
@@ -16,10 +16,20 @@ use super::{
...
 ///
 /// # Example
+/// In this example the genome is an `i32`. `scorer_1` computes the (absolute)
+/// distance between an individual and the constant `11`, `scorer_2` does the
```
NicMcPhee:
```suggestion
/// distance between an individual and the constant `11`, and `scorer_2` does the
```

packages/ec-core/src/individual/ec.rs
```diff
@@ -68,8 +78,8 @@ impl<G, R> Individual for EcIndividual<G, R> {
     /// #
     /// let scorer = FnScorer(|x: &i32| 21_i32.abs_diff(*x));
     ///
-    /// let individual = EcIndividual::new(5, [scorer.score(&5)]);
-    /// assert_eq!(individual.test_results(), &[16]);
+    /// let individual = EcIndividual::new(5, scorer.score(&5));
+    /// assert_eq!(individual.test_results(), &16);
```
NicMcPhee, line 82:
```suggestion
    /// assert_eq!(individual.test_results(), &16);
    /// assert_eq!(individual.test_results(), &0);
```

NicMcPhee: And similar below.

packages/ec-core/src/individual/ec.rs
```diff
@@ -289,6 +301,17 @@ pub trait WithScorer {
 }
 
 impl<GD> WithScorer for GD {
+    /// Create a [`Distribution`] of scored genomes by scoring them using the
+    /// provided `scorer`, turning them into individuals.
```
NicMcPhee:
```suggestion
    /// Create a [`Distribution`] of `EcIndividual`s (i.e., scored genomes)
    /// from a distribution `GD` of genomes. The genomes are scored using the
    /// provided `scorer`, turning them into individuals.
```

packages/ec-core/src/individual/ec.rs
```diff
@@ -324,10 +347,10 @@ where
...
     /// let individual_distribution = StandardUniform.with_scorer_fn(|x: &i32| 100_i32.abs_diff(*x));
     ///
-    /// let my_individual: EcIndividual<_, _> = individual_distribution.sample(&mut rand::rng());
-    ///
-    /// assert!((i32::MIN..=i32::MAX).contains(my_individual.genome()));
-    /// assert!((u32::MIN..=u32::MAX).contains(my_individual.test_results()));
+    /// let my_individual: EcIndividual<i32, u32> = individual_distribution.sample(&mut rand::rng());
+    /// #
+    /// # assert!((i32::MIN..=i32::MAX).contains(my_individual.genome()));
+    /// # assert!((u32::MIN..=u32::MAX).contains(my_individual.test_results()));
```
NicMcPhee, line 354: Why would we make these invisible? I'm not sure I object, but I also don't really understand why we would do this.

packages/ec-core/src/individual/mod.rs
```diff
@@ -1,11 +1,133 @@
 pub mod ec;
 pub mod scorer;
 
+/// A individual in the ec system
+///
+/// This consists of a genome describing the individual and test results for
+/// evaluating the Individual
```
NicMcPhee:
```suggestion
/// An interface representing an evaluated member of the population.
///
/// Types implementing this trait must provide access to:
/// * A [`Genome`](Self::Genome) that defines the individual's encoded traits.
/// * The [`TestResults`](Self::TestResults) that score the individual's performance, error, or quality.
```

packages/ec-core/src/individual/mod.rs
```diff
@@ -1,11 +1,133 @@
...
+///
+/// Also see [`EcIndividual`](ec::EcIndividual)
+///
+/// # Example[^ec-linear-usage]
```
NicMcPhee, line 16: If we wanted to avoid this dependency on `ec-linear` there are a few plausible options:

- Have the genome just be an int (like `u32`) and have the score either be `num_ones()` or distance from a target value a la your examples.
- Have the genome be a `String` and have the score/error be some property of the string. That could be something simple like the Levenshtein distance to a target string, or the absolute difference between the number of digits and non-digits, or the length of the longest ordered prefix.
- Have the genome be a number (maybe a float?) or a tuple of numbers, and then have the error be |f(x)| for some function f.
- Etc.

packages/ec-core/src/individual/mod.rs
```diff
@@ -1,11 +1,133 @@
...
+    /// #         &self.genome
+    /// #     }
+    /// #
+    /// #     fn test_results(&self) -> &ScoreValue<u32> {
+    /// #         &self.score
+    /// #     }
+    /// # }
+    /// #
+    /// let my_individual = ScoredIndividual {
+    ///     genome: [true, false, true].into_iter().collect::<Bitstring>(),
```
NicMcPhee: Given that we repeat the construction of this `BitString`, should we do it once and give it a name?

packages/ec-core/src/individual/scorer.rs
```diff
@@ -68,6 +70,16 @@ where
 
     /// Calculate the score for the passed genome using this [`FnScorer`]'s
     /// function.
+    ///
+    /// # Example
+    /// ```
+    /// # use ec_core::{performance::error_value::ErrorValue, individual::scorer::{Scorer, FnScorer}};
+    /// #
+    /// let scorer = FnScorer(|genome: &i32| ErrorValue(-*genome));
```
NicMcPhee:
```suggestion
    /// let scorer = FnScorer(|&genome: &i32| ErrorValue(-genome));
```

I think this makes the body of the closure easier to read.

packages/ec-core/src/operator/composable/and.rs
```diff
@@ -56,12 +56,20 @@ impl<F, G> And<F, G> {
...
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
 pub enum AndError<T, U> {
+    /// The first operator failed with Error `T`
     First(T),
+    /// The second operator failed with Error `T`
     Second(U),
 }
 
+impl<T: Default, U> Default for AndError<T, U> {
```
NicMcPhee, line 70: I think it would be useful to have a comment here that explains why we implement `Default`, and why we implement this particular way. Even if the implementation choice is somewhat arbitrary (and it is), we should probably say that out loud.

packages/ec-core/src/operator/composable/map.rs
```diff
@@ -78,18 +78,17 @@ where
     }
 }
 
-// I think I can parameterize over the 2 here to make this more general?
-impl<F, Input> Operator<[Input; 2]> for Map<F>
+impl<F, Input, const N: usize> Operator<[Input; N]> for Map<F>
```
NicMcPhee, line 81: Nice – thanks.

packages/ec-core/src/operator/composable/map.rs
```diff
@@ -99,14 +98,30 @@ where
...
     fn apply<R: Rng + ?Sized>(
         &self,
-        [x, y]: [Input; 2],
+        input: [Input; N],
         rng: &mut R,
     ) -> Result<Self::Output, Self::Error> {
-        let first_result = self.f.apply(x, rng).map_err(|e| MapError(e, 0))?;
-        let second_result = self.f.apply(y, rng).map_err(|e| MapError(e, 1))?;
-        Ok([first_result, second_result])
+        // FIXME: This could be made more efficent by using MaybeUninit<T> over
```
NicMcPhee:
```suggestion
        // FIXME: This could be made more efficient by using MaybeUninit<T> over
```

NicMcPhee: I'm not sure how I feel about "FIXME" as the label on these ideas for introducing `unsafe` features like `MaybeUninit`. `FIXME` implies that something is _broken_ and that we ought to "fix" it. My inclination is that something would be need to be pretty seriously inefficient to warrant introducing `unsafe` code, so I'm not sure this is the message we want to send.

packages/ec-core/src/operator/composable/mod.rs
```diff
@@ -5,54 +5,255 @@ mod then;
...
+#[doc(hidden)]
 pub use ec_macros::Composable;
 pub use map::*;
 pub use repeat_with::*;
 pub use then::*;
 
+/// Create new [`Operator`]'s by composing preexisting [`Operator`]'s!
+///
+/// This is usually the prefered method for interacting with operators,
+/// similarly how one would use composable iterators over defining new iterator
```
NicMcPhee:
```suggestion
/// similarly to how one would use composable iterators over defining new iterator
```

packages/ec-core/src/operator/composable/mod.rs
```diff
@@ -5,54 +5,255 @@ mod then;
...
+/// types all the time.
+///
+/// # Example
```
NicMcPhee, line 20:
```suggestion
/// # Example

/// In the example below:
///    - `Constant::new(false)` creates an operator which returns `false` regardless of which value it is applied to.
///    - `.apply_n_times::<5>()` applies the previous operator (`Constant`) five times to the input, generating an array of 5 instances of `false`.
///    - `.then_map(Identity)` maps the `Identity` operator over each of those instances of `false`, returning another array of 5 instances of `false`.
///.   - `.then_map(Constant::new(5))` converts each of those `false` values into `5`, yielding an array of 5 instance of the value 5.
```

NicMcPhee, line 20: There's a lot I like about this example, but I do have two questions.

First, nothing in this example _explicitly_ uses `Composable`; it's all tucked away in things like `then_map()`. Should we either explicitly construct a `Composable`, or add a comment that highlights the fact the `then_map()` is doing the composition?

Second, using the `Identity` operator doesn't actually _do_ thing. Would it make more sense to have an operator like `Double` or something that _does_ something? 

There are similar comments in several places below. One thing that all this suggests (and which you have addressed some in #418) is that there are things here that just don't make a lot of sense and should perhaps be removed/modified. We probably don't want to spend a ton of time documenting things we wish to promptly remove, but I don't think that's everything here?

packages/ec-core/src/operator/composable/mod.rs
```diff
@@ -5,54 +5,255 @@ mod then;
...
+/// ```
+/// # use ec_core::operator::{constant::Constant, identity::Identity, Composable, Operator};
+/// # use rand::rng;
+/// #
+/// let my_operator = Constant::new(false)
+///     .apply_n_times::<5>()
```
NicMcPhee, line 39: I would definitely avoid using `5` as both the value and the length here. It's easy to misread `[5; 5]` as `[5, 5]`, which is a totally different thing.

Again, though, I think the use of `Identity` and `Constant::new(5))` isn't optimal, and maybe using an operator that "does something" would be better than just a lot of copying and overwriting.

packages/ec-core/src/operator/composable/mod.rs
```diff
@@ -5,54 +5,255 @@ mod then;
...
+    /// ```
+    /// # use ec_core::operator::{
+    /// #     constant::Constant,
+    /// #     composable::And,
+    /// #     Composable,
+    /// #     Operator
+    /// # };
+    /// # use rand::rng;
+    /// #
+    /// let my_composed_operator = Constant::new(5).and(Constant::new(6));
```
NicMcPhee, line 123: I'm not sold on the names (e.g.) `my_composed_operator` since they are not actually operators but (in this case) pairs. The `assert_eq!` line below that compares them really wouldn't work if they were truly operators (functions).

This applies elsewhere in the file.

packages/ec-core/src/operator/composable/mod.rs
```diff
@@ -5,54 +5,255 @@ mod then;
...
+    /// #     },
+    /// #     individual::{
+    /// #         scorer::FnScorer,
+    /// #         ec::EcIndividual,
+    /// #     }
+    /// # };
+    /// # use rand::rng;
+    /// #
+    /// let my_individual_generator =
+    ///     Constant::new(5).wrap::<GenomeScorer<_, _>>(FnScorer(|x: &i32| *x));
```
NicMcPhee: Again, using identity here (`|x| x`) isn't terribly informative. If we just changed that to `x*2` or `x+7` or something, then the assertion would (I think) be clearer (`(5, 10)` or `(5, 12)`).

Ditto below.

packages/ec-core/src/operator/composable/repeat_with.rs
```diff
@@ -29,12 +29,21 @@ pub struct RepeatWith<F, const N: usize> {
...
     /// [`RepeatWith`]. This is identical to [`RepeatWith::new`]
     ///
-    /// # Example
+    /// # Examples
+    /// ```
+    /// # use ec_core::operator::{Operator, constant::Constant, composable::RepeatWith};
+    /// # use std::convert::Infallible;
+    /// let repeated_operator = RepeatWith::<_, 5>::from(Constant::new(1));
+    /// #
+    /// # let value = repeated_operator.apply((), &mut rand::rng())?;
```
NicMcPhee, line 43: I think I would uncomment the `let value = ` and `assert_eq!()` lines so people can see what this _does_ in the documentation.

packages/ec-core/src/operator/composable/then.rs
```diff
@@ -30,6 +30,48 @@ pub struct Then<F, G> {
...
+    ///
+    /// # Examples
+    /// ```
+    /// # use ec_core::operator::{
+    /// #     Operator,
+    /// #     composable::{Map, Then, RepeatWith},
+    /// #     constant::Constant,
+    /// #     identity::Identity
+    /// # };
+    /// #
```
NicMcPhee, line 46: - It would be nice to use something other than `Identity`.
- It's not clear why we have two examples here or how they relate.
- I think it would be useful to not comment out the assertions so people can see what these do.

packages/ec-core/src/operator/mutator/erased.rs
```diff
@@ -49,11 +49,60 @@ use super::Mutator;
...
+    /// ```
+    /// # use rand::{rng, Rng};
+    /// # use ec_core::operator::mutator::{Mutator, DynMutator};
+    /// # use std::convert::Infallible;
+    /// #
+    /// type Genome<T> = [T; 4];
+    /// #
+    /// # struct FlipOne;
+    /// #
+    /// # impl Mutator<Genome<bool>> for FlipOne {
```
NicMcPhee, line 76: We implement `FlipOne` and similar things in tests many times. Would it make sense to have a `demos` module in `ec-core/src` that has some simple reference implementations like `FlipOne`? These could then be used as examples in "book-like" documentation and as running examples in doctests.

We don't want to delay this PR for this idea, but I wanted to throw it out there for discussion.

packages/ec-core/src/operator/mutator/erased.rs
```diff
@@ -49,11 +49,60 @@ use super::Mutator;
...
+    /// #         genome[index] = !genome[index];
+    /// #         Ok(genome)
+    /// #     }
+    /// # }
+    ///
+    /// let my_erased_mutator: Box<dyn DynMutator<Genome<bool>>> = Box::new(FlipOne);
+    ///
+    /// let genome = [true, false, false, true];
+    /// let child_genome = (*my_erased_mutator).dyn_mutate(genome, &mut rng()).unwrap();
+    /// # let num_diffs = genome
```
NicMcPhee, line 95: If we're going to "hide" the "interesting" parts of the tests, would it make sense to have an English language comment that summarizes the expected behavior? Something like:

"`child_genome` should differ from `genome` in example one location."

packages/ec-core/src/operator/mutator/mod.rs
```diff
@@ -250,7 +285,7 @@ pub trait Mutator<G> {
 /// assert_eq!(chain.apply(genome, &mut rng())?, 1);
 /// # Ok::<(), Box<dyn std::error::Error>>(())
 /// ```
-#[derive(Debug, Composable, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Composable, Default)]
```
NicMcPhee: Adding `Copy` requires that `M` be `Copy`, correct? Do we want/need to require that? What if a mutator wrapped a `Vec` of values, for example?

packages/ec-core/src/operator/recombinator/mod.rs
```diff
@@ -95,11 +95,44 @@ pub trait Recombinator<GS> {
...
+    /// #         let index = rng.random_range(0..first_parent.len());
+    /// #         first_parent[index] = second_parent[index];
+    /// #         Ok(first_parent)
+    /// #     }
+    /// # }
     ///
-    /// This will return an error if there is an error recombining the given
-    /// parent genomes. This will usually be because the given `genomes` are
-    /// invalid in some way, thus making recombination impossible.
+    /// let first_parent = [0, 0, 0, 0];
```
NicMcPhee, line 123: Thinking about it more as I read through, it seems weird to show people the explicit values *before* recombination (or mutation), but then not say anything about the value _after_ the operator is applied. It does seem that we want to either expose the assertions (my preference) or have a short English-language description of what should happen.

packages/ec-core/src/operator/selector/best.rs
```diff
@@ -28,7 +28,19 @@ where
 {
     type Error = EmptyPopulation;
 
-    /// Select an Individual from the given Population using this selector.
+    /// Select the individual with the highest value.
```
NicMcPhee:
```suggestion
    /// Select an individual with the "best" value (e.g., high for scores, low for errors). If there are multiple individuals with the same values, the one will be returned at random.
```

packages/ec-core/src/operator/selector/dyn_weighted.rs
```diff
@@ -18,6 +18,21 @@ use crate::population::Population;
...
 /// and actual selection will then be forwarded to that selector.
+///
+/// # Example
+/// ```
+/// # use ec_core::operator::selector::{best::Best, worst::Worst, dyn_weighted::{DynWeighted, DynWeightedError}, Selector};
+/// # use rand::rng;
+/// #
+/// type Population = Vec<i32>;
+///
+/// let dyn_selector = DynWeighted::<Population>::new(Best, 1).with_selector(Worst, 1);
```
NicMcPhee, line 30: I think we need a comment on this line, e.g.,

```rust
// Select the best or the worst individual with equal likelihood (because the two selectors both have the same weights, namely 1).
```

Probably the same thing on the next example.

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -57,13 +57,38 @@ impl Lexicase {
...
     /// selection might fail.
+    ///
+    /// # Example
+    /// ```
+    /// # use ec_core::{
+    /// #     individual::ec::EcIndividual,
+    /// #     operator::selector::{Selector, lexicase::Lexicase},
+    /// #     performance::test_results::TestResults
+    /// # };
+    /// let population = [
```
NicMcPhee, line 76:
```suggestion
    /// // The first individual will always be
    /// // selected because both its test case
    /// // scores are as high or higher than for
    /// // all the other individuals.
    /// let population = [
```

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -57,13 +57,38 @@ impl Lexicase {
...
+    ///     EcIndividual::new(100, TestResults::<i32>::from_iter([4, 10])),
+    ///     EcIndividual::new(100, TestResults::<i32>::from_iter([2, 10])),
+    ///     EcIndividual::new(100, TestResults::<i32>::from_iter([4, 5])),
+    ///     EcIndividual::new(100, TestResults::<i32>::from_iter([2, 5])),
+    /// ];
+    ///
+    /// let lexicase = Lexicase::new(2);
+    ///
+    /// # let selected = lexicase.select(&population, &mut rand::rng())?;
```
NicMcPhee, line 85: I really don't see why we'd "hide" these lines? At a minimum I think we should expose the `let selected =` line, since I think that would be more consistent with other examples.

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -57,13 +57,38 @@ impl Lexicase {
...
     #[must_use]
     pub const fn new(num_test_cases: usize) -> Self {
         Self { num_test_cases }
     }
 }
 
-/// Error that can occur during [`Lexicase`] selection:
+/// Errors that can occur during [`Lexicase`] selection:
 /// - [`EmptyPopulation`] when trying to select from an empty population, or
 /// - `MissingTestCase` when using a lexicase selector with a higher number of
```
NicMcPhee, line 101: Why don't we have `MissingTestCase` linking to somewhere?

packages/ec-core/src/operator/selector/lexicase.rs
```diff
@@ -107,6 +132,31 @@ where
...
     /// Select an Individual from the given Population using this selector.
     ///
+    /// # Example
+    /// ```
+    /// # use ec_core::{
+    /// #     individual::ec::EcIndividual,
+    /// #     operator::selector::{Selector, lexicase::Lexicase},
+    /// #     performance::test_results::TestResults
+    /// # };
+    /// let population = [
```
NicMcPhee, line 154: Repeat the added comment from above about the first always being selected.

packages/ec-core/src/operator/selector/mod.rs
```diff
@@ -253,7 +265,7 @@ where
 /// assert_eq!(choice_length, 5);
 /// # Ok::<(), Box<dyn std::error::Error>>(())
 /// ```
-#[derive(Debug, Clone, Composable, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
+#[derive(Debug, Copy, Clone, Composable, PartialEq, Eq, PartialOrd, Ord, Hash, Default)]
```
NicMcPhee: Do we *need* `Copy` here? Could that constrain us later?

packages/ec-core/src/operator/selector/tournament.rs
```diff
@@ -96,26 +111,50 @@ impl TournamentSizeError {
...
 
 impl Tournament {
-    /// Construct a tournament selector with the given tournament size. This
-    /// will select `size` individuals from the population, randomly
+    /// Construct a tournament selector with the given tournament size.
+    ///
+    /// This will select `size` individuals from the population, randomly
     /// without replacement, and return the "best" from that set.
+    ///
+    /// If the size you construct from is a constant, consider using
```
NicMcPhee, line 119: Can we add our own Clippy warnings in some way? This would be a perfect case where a Clippy-style warning would be really useful.

packages/ec-core/src/operator/selector/tournament.rs
```diff
@@ -33,12 +37,23 @@ use crate::population::Population;
 /// assert!([2, 5, 8].contains(selected));
 /// # Ok::<(), TournamentSizeError>(())
 /// ```
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Tournament {
```
NicMcPhee, line 41: We never say anywhere (although it's implied in the last example) that this is tournament *without replacement*. It seems that we should say that out loud, since there are people who implement tournament with replacement.

NicMcPhee (changes requested):
Huzzah! I got through it all!

Thanks so much for all your work on this. I've got a few, mostly minor, suggestions. Hopefully we can wrap this up soon.

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
JustusFluegel: I like `specified`.

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
JustusFluegel: I am not totally sure it buys us much but i am not opposed either so i did do it. I think the point of me using `Vec's` in the first place was to choose something that is not trivially copyable so that we actually see the clone is happening, so i choose owned `String` types.

packages/ec-core/src/distributions/wrappers/choose_cloning.rs
JustusFluegel: I think i like the current approach of having a `let foo = ...` followed by a `# let _ = foo` (the second one being there just for clippy) because in my mind the non-underscored foo signifies to the reader that foo is something they might actually want to use; using a `_foo` instead kind of implies that the value is irrelevant to the example which it most certainly is not.

packages/ec-core/src/distributions/wrappers/choose_cloning_owning.rs
JustusFluegel: I don't think we talked about this before, but the way I see it `Borrow` more directly semantically corresponds to the `&x` operation, whereas `AsRef` is another operation on its own (`x.as_ref()`) which in theory can correspond not to a borrowed variant of `T` but rather to a member of `T` or similar (not directly related to the entire `T`) so i think `Borrow` expresses the intent better.
 

I don't think it's something we should touch in this pr anyways tho, since I don't believe (at least GitHub is not showing it as such above) this was introduced in this pr.


But I am open to having a discussion about this at a later point in time, i am not really that partial to using `Borrow` either.

JustusFluegel: Good catch. I think i inverted a `<=` expression at some point here and made the common error of forgetting to omit the `=` part.

packages/ec-core/src/individual/ec.rs
JustusFluegel: Sounds correct to me.

packages/ec-core/src/individual/ec.rs
JustusFluegel: I don't think I quite understand what you are trying to do here, the second assertion you proposed should always fail, no? `test_results()` only borrows and returns a borrowed value and doesn't modify it as far as i understand.

packages/ec-core/src/individual/ec.rs
JustusFluegel: I think it was mostly in the interest of making the examples really short and easily parseable to the one reading the documentation, not taking up too much space with stuff that is not relevant to the basic usage. The asserts were more of a add-on than actually intended in the first place for the example, just to make it more useful as a doctest.

packages/ec-core/src/operator/selector/tournament.rs
JustusFluegel: I think it's possible but rather more involved (we'd need some kind of `cargo unhindered lint`) but people have certainly done that before using clippy as a basis - not a consideration for this pr specifically tho imo.

Feel free to open an issue.

packages/ec-core/src/operator/selector/lexicase.rs
JustusFluegel: Because we are currently documenting the exact enum its a variant of and there isn't any external struct to link to.

packages/ec-core/src/operator/selector/lexicase.rs
JustusFluegel: Note that this example is explicitly  documenting the  new function. The whole example can be found to the user as part of the main `Lexicase` documentation, and I'd want to avoid repeating the entirety of it on individual methods to avoid clutter.

packages/ec-core/src/individual/mod.rs
JustusFluegel: Sure, but the whole point of using ec-linear was to avoid to need to implement various traits again on the actual type used and to provide a realistic example. I think the dependency on ec-linear is fine in this context.

packages/ec-core/src/individual/ec.rs
JustusFluegel: I think it was mostly in the interest of making the examples really short and easily parseable to the one reading the documentation, not taking up too much space with stuff that is not relevant to the basic usage. The asserts were more of a add-on than actually intended in the first place for the example, just to make it more useful as a doctest.

packages/ec-core/src/operator/composable/map.rs
JustusFluegel: Maybe change the wording to "FIXME: Check if ... is a noticeable performance improvement`?

packages/ec-core/src/operator/composable/mod.rs
JustusFluegel: I'll admit i was kind of lazy but i also wanted to demonstrate the built in operators we have. 

I am not quite sure if i want to change too much as part of this pr. I added a comment about what composable is actually doing here tho.

packages/ec-core/src/operator/composable/mod.rs
JustusFluegel: well they are operators, both are the RepeatWith composed operators (in the sense of the `Operator` trait we introduced). 

Feel free to suggest alternative names tho.

packages/ec-core/src/operator/composable/repeat_with.rs
JustusFluegel: Again like we had above this is documenting the `from` method specifically. The same example is available above as part of the general `RepeatWith` documentation and i want to avoid cluttering the page.

packages/ec-core/src/operator/composable/then.rs
JustusFluegel: The two examples demonstrate from and into respectively. I added two lines above each, to spell that out explicitly. Same argument about the assertions as above.

packages/ec-core/src/operator/mutator/erased.rs
JustusFluegel: Maybe, but not quite sure what the way is to best do this from a organizational perspective.

packages/ec-core/src/operator/recombinator/mod.rs
JustusFluegel: It is on the main example on Recombinator directly rather than the methods.

packages/ec-core/src/operator/selector/lexicase.rs
NicMcPhee: I can buy that. Thanks for (re)explaining it to me.

packages/ec-core/src/individual/mod.rs
```diff
@@ -1,11 +1,137 @@
 pub mod ec;
 pub mod scorer;
 
+/// An interface representing an evaluated member of the population.
+///
+/// Types implemeting this trait must provide access to:
```
NicMcPhee, line 6:
```suggestion
/// Types implementing this trait must provide access to:
```

packages/ec-core/src/individual/mod.rs
```diff
@@ -1,11 +1,137 @@
...
+/// - A [`Genome`](Self::Genome) that defines the individuals encoded traits.
```
NicMcPhee, line 7:
```suggestion
/// - A [`Genome`](Self::Genome) that defines the individual's encoded traits.
```

packages/ec-core/src/operator/composable/and.rs
```diff
@@ -56,12 +56,23 @@ impl<F, G> And<F, G> {
...
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
 pub enum AndError<T, U> {
+    /// The first operator failed with Error `T`
     First(T),
+    /// The second operator failed with Error `T`
     Second(U),
 }
 
+// The implementation of Default here is somewhat arbitrary and more for
+// actually having a value in i.e. a global state if it is required rather than
```
NicMcPhee, line 68:
```suggestion
// actually having a value in, e.g., a global state if it is required rather than
```

packages/ec-core/src/operator/composable/mod.rs
```diff
@@ -5,54 +5,268 @@ mod then;
...
+/// # Example
+/// In the example below:
+///    - `Constant::new(false)` creates an operator which returns `false`
+///      regardless of which value it is applied to.
+///    - `.apply_n_times::<5>()` applies the previous operator (`Constant`) five
+///      times to the input, generating an array of 5 instances of `false`.
+///    - `.then_map(Identity)` maps the `Identity` operator over each of those
+///      instances of `false`, returning another array of 5 instances of
+///      `false`.
+///    - `.then_map(Constant::new(5))` converts each of those `false` values
```
NicMcPhee, line 29:
```suggestion
///    - `.then_map(Constant::new(10))` converts each of those `false` values
```

packages/ec-core/src/operator/selector/best.rs
```diff
@@ -28,7 +28,26 @@ where
 {
     type Error = EmptyPopulation;
 
-    /// Select an Individual from the given Population using this selector.
+    /// Select an individual with the "best" value.
+    ///
+    /// The "best" value is determined by the [`Ord`] implementation on the
+    /// `Individual`, i.e. if you are using `ErrorValue` and `ScoreValue` high
+    /// for scores and low for errors.
```
NicMcPhee, line 35:
```suggestion
    /// `Individual`. If, for example, you are using `ErrorValue` then low values will be "best". If, on the other hand, you are using `ScoreValue`, then high values will be the "best".
```

packages/ec-core/src/operator/selector/best.rs
```diff
@@ -28,7 +28,26 @@ where
...
+    ///
+    /// If there is no single best value it is unspecified which of them is
```
NicMcPhee, line 37:
```suggestion
    /// If there is no single best value it is unspecified which of those tied for best is
```

packages/ec-core/src/operator/selector/tournament.rs
```diff
@@ -6,14 +6,16 @@ use rand::{Rng, prelude::IndexedRandom};
 use super::Selector;
 use crate::population::Population;
 
-/// Tournament selector.
+/// Tournament without replacement selector.
```
NicMcPhee, line 9:
```suggestion
/// Tournament selection without replacement.
```

NicMcPhee (changes requested):
A few minor suggestions from the stream.

Huge thanks for all the work!

NicMcPhee (approved):
Thanks for all this work!

PR #443, opened 2026-02-03
Use while let instead of loop { let ... else { break; } .. }

JustusFluegel:
Fixes new lint in nightly ci


JustusFluegel:
> I've added a suggestion to further refactor this loop to be more idiomatic and efficient by avoiding costly Vec::remove calls inside the loop.

what do you mean, gemini? :joy:

PR #457, opened 2026-03-01
Provide default summation overflow handling

### LINKED ISSUE #20: Handle overflow in total error

NicMcPhee:
Defines traits `DefaultAccumulator` and `AccumulateInto` to specify how various numeric types should behave when summed, primarily to mitigate potential overflows.

Signed integers are promoted to a wider type to prevent overflow during accumulation. Unsigned integers are handled using saturating arithmetic, which clamps sums at their maximum value. This provides predictable overflow behavior and enhances data integrity in performance measurements.

See #20 and [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2f7aaab46c7427fda8409fa21a525ca0) for additional details.

Closes #20


JustusFluegel:
Hey Nic thanks for the work but I think esitsu convinced me that we should maybe design the traits somewhat differently; maybe that is something we should discuss this stream and then we can revisit this pr?

JustusFluegel:
@NicMcPhee do you remember what was going on here? Is this still required?

PR #479, opened 2026-08-09
Fix unnecessary double reference; use `assert_ne!`

### LINKED ISSUE #478: Remove unnecessary reference in `format!` call
In `push-macros/src/doctest_tokenstream.rs` we have:

```rust
   .map(|l| format!(" {line}", line = &l.get(4..).unwrap_or_default()))
```

which Clippy will no longer accept. [The problem is that the `&` isn't necessary](https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#useless_borrows_in_formatting) since `format!` already takes a reference. Thus adding `&` creates an unnecessary double reference, which is apparently some tiny slower.

NicMcPhee:
Clippy was unhappy about a double reference (see #478), which this fixes. 

We also fixed a Clippy warning about using `assert!(x != y)` when we should use `assert_ne!(x, y)`.

Closes #478

### DIFF
--- a/packages/ec-core/src/performance/test_result.rs
+++ b/packages/ec-core/src/performance/test_result.rs
@@ -171,7 +171,7 @@ pub(crate) mod test_result_tests {
         let first: TestResult<i32, i32> = TestResult::Score(ScoreValue(32));
         let second = TestResult::Score(ScoreValue(87));
         assert!(first < second);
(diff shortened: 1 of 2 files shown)


JustusFluegel (approved):
LGTM

PR #484, opened 2026-08-16
Bump rand and fix errors

JustusFluegel:
`RngCore` was renamed to `Rng` and `Rng` was renamed to `RngExt` so the `dyn RngCore` types became `dyn Rng` and we needed a bunch of `RngExt` imports.


NicMcPhee (approved):
Thanks!

PR #488, opened 2026-08-29
Refactor input variable representation

### LINKED ISSUE #172: Don't panic on with_input() with invalid variable
[This](https://github.com/JustusFluegel/rust-ga/blob/8576fd7da643fd30f1e64b33d51b29a2afb7b063/packages/push/src/push_vm/push_state.rs#L86C17-L89C18) currently panics. We should deal with this better.

### LINKED ISSUE #487: Refactor input variable representation
The old way of handling input variables isn't consistent with how other instructions are handled. This should be fixed in order to help "automate" (via macros) the creation of new states.

NicMcPhee:
Makes the handling of input variables more uniform, and easier to "automate" via macros.

We also introduced `UnknownVariableError` and `WithInputInstructionError` to avoid panicking if there is an unknown variable. A downside of the current arrangement is that if there is an unknown variable, the system appears to run successfully (with your score being a penalty value) even though nothing useful is happening.

Closes #487 
Closes #172


packages/push/src/instruction/with_input/error.rs
```diff
@@ -0,0 +1,88 @@
...
+    instruction::instruction_error::PushInstructionError, push_vm::variables::UnknownVariableError,
+};
+
+#[derive(thiserror::Error, Debug, Clone, Eq, PartialEq, PartialOrd, Ord)]
+pub enum WithInputInstructionError<T> {
+    #[error(transparent)]
+    Instruction(T),
+    #[error(transparent)]
+    UnknownVariable(#[from] UnknownVariableError),
+}
```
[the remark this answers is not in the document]
NicMcPhee: This is actually not a problem.

packages/push/src/instruction/with_input/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    /// # Errors
+    ///
+    /// - Returns an error if performing the input instruction associated with
+    ///   this variable returns an error.
+    /// - Returns a fatal [`UnknownVariableError`] if there is an attempt to
+    ///   lookup the value of a variable that isn't registered with the state.
+    fn perform(&self, state: S) -> InstructionResult<S, Self::Error> {
+        let Some(instruction) = state.get_input_instruction(&self.0) else {
+            return Err(Error::fatal(state, UnknownVariableError(self.0.clone())));
```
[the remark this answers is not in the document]
NicMcPhee: This is a reasonable point and we should probably add a simple test for this.

packages/push/src/error/mod.rs
```diff
@@ -27,7 +27,18 @@ pub trait MapInstructionError<S, E> {
     /// # Errors
     ///
     /// This always returns an error type.
-    fn map_err_into(self) -> InstructionResult<S, E>;
+    fn map_err_into<E2>(self) -> InstructionResult<S, E2>
+    where
+        Self: Sized,
+        E: Into<E2>;
+
+    ///
```
JustusFluegel, line 47: I know that this is partially my fault but maybe we should actually write something here.

packages/push/src/instruction/mod.rs
```diff
@@ -100,20 +101,33 @@ impl PushInstruction {
...
-            }
-            Self::Exec(i) => i.perform(state),
-            Self::BoolInstruction(i) => i.perform(state),
-            Self::IntInstruction(i) => i.perform(state),
-            Self::FloatInstruction(i) => i.perform(state),
+            // Self::InputVar(var_name) => {
+            //     // TODO: Should `push_input` return the new state?
+            //     // Or add a `with_input` that returns the new state and keep `push_input`?
+            //     state.with_input(var_name)
+            // }
```
JustusFluegel: Should we just delete this?

packages/push/src/instruction/mod.rs
```diff
@@ -122,6 +136,36 @@ impl Instruction<PushState> for PushInstruction {
...
+//             Self::IntInstruction(i) => i.perform(state),
+//             Self::FloatInstruction(i) => i.perform(state),
+//             Self::PrintString(i) => i.perform(state).map_err_into(),
+//             Self::PrintSpace(i) => i.perform(state).map_err_into(),
+//             Self::PrintNewline(i) => i.perform(state).map_err_into(),
+//             Self::PrintPeriod(i) => i.perform(state).map_err_into(),
+//         }
+//     }
+// }
+
```
JustusFluegel: What is this still doing here?

packages/push/src/instruction/mod.rs
```diff
@@ -142,7 +186,7 @@ impl NumOpens for PushInstruction {
 impl std::fmt::Display for PushInstruction {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Self::InputVar(instruction) => write!(f, "{instruction}"),
+            Self::WithInput(instruction) => write!(f, "{instruction}"),
```
JustusFluegel:
```suggestion
            Self::WithInput(instruction) => instruction.fmt(f),
```

instruction.fmt(f) is cleaner instead of write if we don't want to add anything

packages/push/src/push_vm/push_state.rs
```diff
@@ -146,6 +101,14 @@ impl State for PushState {
     }
 }
 
+impl HasInputs for PushState {
+    type InputInstruction = ConstantExpression;
+
+    fn get_input_instruction(&self, var_name: &VariableName) -> Option<Self::InputInstruction> {
```
JustusFluegel, line 111: Do we want some documentation on this like we previously had on the `with_input` this pr now removes in favor of this method?

JustusFluegel (changes requested):
Some minor nitpicks, especially about removing commented out code that is no longer required after this pr.

JustusFluegel (approved):

